PR #10, opened 2019-01-09
Lambda tweaks

Mullefa:
This PR resolves a couple of issues with the current lambda:
- stops environment variables from being wiped on each deploy by defining them in cloud formation (good spot @Calum-Campbell)
- uses logback and a custom appender (from jlib-awslambda-logback) to ensure that multi-line log entries are still just one event in logback
- derives the config and initialise the services from within the `Lambda::handler()` function (to try and provide more meaningful log messages if the environment variables required by the configuration are missing)

### DIFF
--- a/payment-failure/build.sbt
+++ b/payment-failure/build.sbt
@@ -6,18 +6,19 @@ scalaVersion := "2.12.6"
 val circeVersion = "0.10.1"
 
 libraryDependencies ++= Seq(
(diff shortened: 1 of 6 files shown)


Calum-Campbell (approved):

Mullefa:
@Calum-Campbell - before merging I'll update the PROD cloud formation via the UI (as there are parameter changes)

PR #17, opened 2019-01-22
add advice on re-handling messages on the dead letter queue

Mullefa:
Note: the advice here doesn't just apply to the payment failure lambda; therefore, it could be moved in future (to somewhere more general). For now, put it here, as we have needed to do this a couple of times for this lambda already.


jfsoul:
Am I right in thinking there are some scenarios where we would not want to try moving the message back to the main queue? Might be good to add a mention that moving it back is not always the best course of action (e.g. when the email address fails validation, it will keep failing).

jfsoul (approved):

PR #18, opened 2019-01-22
Accumulate errors using validated

Mullefa:
Here's a (possibly) controversial PR 😉

We've done (IMO) a nice job of handling errors in this lambda using the `Either[Throwable, ?]` monad. Let's leverage this to accumulate the errors when handling a batch of messages.

The main advantage (apart from it being more idiomatic with this style of Scala - not sure if that counts), is that we accumulate the errors and then throw them all at once (so no information is lost), rather than throwing on the first error encountered.

I've also added a few more comments explaining what throwing an error means in the context of a lambda with a queue as its event source.

### DIFF
--- a/payment-failure/src/main/scala/com/gu/identity/paymentfailure/LambdaService.scala
+++ b/payment-failure/src/main/scala/com/gu/identity/paymentfailure/LambdaService.scala
@@ -9,7 +9,7 @@ import org.slf4j.MDC
 
 import scala.collection.JavaConverters._
 
-class LambdaService private (sqsService: SqsService, sendEmailService: SendEmailService) {
+class LambdaService(sqsService: SqsService, sendEmailService: SendEmailService) {
 
   def processMessage(message: SQSMessage): Either[Throwable, BrazeResponse] =
     for {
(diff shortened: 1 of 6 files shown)


payment-failure/src/main/scala/com/gu/identity/paymentfailure/LambdaService.scala
```diff
@@ -1,22 +1,34 @@
...
-        _ <- sqsService.processDeleteMessageResult(result)
-      } yield brazeResponse
-    }
+class LambdaService(sqsService: SqsService, sendEmailService: SendEmailService) {
+
+  def processMessage(message: SQSMessage): Either[Throwable, BrazeResponse] =
+    for {
+      emailData <- sqsService.parseSingleMessage(message)
+      brazeResponse <- sendEmailService.sendEmail(emailData)
+      // Deleting a message from the queue will mean that even if the lambda throws an error
```
Calum-Campbell, line 18: nice

Calum-Campbell (approved):
Looks good! Could we also validate getting the config in a similar way? No need to in this PR just a thought.

Mullefa:
@Calum-Campbell re config - exactly - that's a perfect use case - in fact it's even mentioned in the [Validated docs](https://typelevel.org/cats/datatypes/validated.html)! Roughly speaking `Either` is for any stuff that needs to be done sequentially; `Validated` is for stuff that can be done in parallel.

PR #19, opened 2019-01-23
Still send email if email encryption fails

Mullefa:
Currently we don't send a payment failure email if email encryption fails. Change this so that the email is still sent: preferable an email with no encrypted email token than none at all. 

Note, this is not a regression in that it was the current behaviour when emails were sent by membership-workflow.

The (single) cause of email encryption failure witnessed in PROD so far is that the email address is in on a blocked domain. One fix for this would be to allow the encryption of emails on blocked domains.

cc @jacobwinch

### DIFF
--- a/payment-failure/src/main/scala/com/gu/identity/paymentfailure/BrazeClient.scala
+++ b/payment-failure/src/main/scala/com/gu/identity/paymentfailure/BrazeClient.scala
@@ -2,31 +2,32 @@ package com.gu.identity.paymentfailure
 
 import com.typesafe.scalalogging.StrictLogging
 import cats.syntax.either._
+import io.circe.parser.decode
 import io.circe.syntax._
 import scalaj.http.Http
 
 
 class BrazeClient(config: Config) extends StrictLogging {
 
-  def sendEmail(emailData: IdentityBrazeEmailData, emailToken: String) : Either[Throwable, BrazeResponse] = {
+  def sendEmail(request: BrazeSendRequest) : Either[Throwable, BrazeResponse] = {
 
-    logger.info(s"sending email via Braze - email data: $emailData")
-
-    val sendRequest = BrazeSendRequest(config.brazeApiKey, emailData.templateId, List(BrazeRecipient(emailData.externalId, emailData.customFields + ("emailToken" -> emailToken))))
+    logger.info(s"sending email via Braze - request data: $request")
 
     Either.catchNonFatal(
       Http(s"${config.brazeApiHost}/campaigns/trigger/send")
       .header("content-type", "application/json")
-      .postData(sendRequest.asJson.toString)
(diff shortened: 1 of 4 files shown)


payment-failure/src/main/scala/com/gu/identity/paymentfailure/BrazeClient.scala
```diff
@@ -2,31 +2,32 @@ package com.gu.identity.paymentfailure
...
-    logger.info(s"sending email via Braze - email data: $emailData")
-
-    val sendRequest = BrazeSendRequest(config.brazeApiKey, emailData.templateId, List(BrazeRecipient(emailData.externalId, emailData.customFields + ("emailToken" -> emailToken))))
+    logger.info(s"sending email via Braze - request data: $request")
 
     Either.catchNonFatal(
       Http(s"${config.brazeApiHost}/campaigns/trigger/send")
       .header("content-type", "application/json")
-      .postData(sendRequest.asJson.toString)
+      .postData(request.asJson.toString)
```
Mullefa, line 19: shorten to `request` - send request context not needed

payment-failure/src/main/scala/com/gu/identity/paymentfailure/BrazeClient.scala
```diff
@@ -2,31 +2,32 @@ package com.gu.identity.paymentfailure
...
 import cats.syntax.either._
+import io.circe.parser.decode
 import io.circe.syntax._
 import scalaj.http.Http
 
 
 class BrazeClient(config: Config) extends StrictLogging {
 
-  def sendEmail(emailData: IdentityBrazeEmailData, emailToken: String) : Either[Throwable, BrazeResponse] = {
+  def sendEmail(request: BrazeSendRequest) : Either[Throwable, BrazeResponse] = {
```
Mullefa, line 12: Have the email service be responsible for constructing a Braze request; the client is just concerned about executing it.

payment-failure/src/main/scala/com/gu/identity/paymentfailure/SendEmailService.scala
```diff
@@ -1,11 +1,35 @@
 package com.gu.identity.paymentfailure
 
-class SendEmailService (identityClient: IdentityClient, brazeClient: BrazeClient, config: Config){
+import com.typesafe.scalalogging.StrictLogging
+
+class SendEmailService(identityClient: IdentityClient, brazeClient: BrazeClient, config: Config) extends StrictLogging {
+
+  // Recover from error by allowing for the token to be optional.
+  // Would rather send an email without a token than not at all.
+  def encryptEmail(email: String): Option[String] =
```
Mullefa, line 9: The essence of the PR

laurenemms (approved):

PR #22, opened 2019-02-22
Generate an AB test to test encrypted email token

Mullefa:
This PR has quite a big diff (relatively). I decided to make the `SendEmailService` (renamed to `BrazeEmailService`) a trait to allow for different implementations. This means that the default email service can be (temporarily) be swapped out for other services (in this instance to run the encrypted email token test). What this means in future is that similar changes won't touch too many lines of pre-existing code; just the writing of new code. In addition to making subsequent PRs easier to review, it also means that the tests relating to the email service don't have to be re-written on every change.

### DIFF
--- a/payment-failure/src/main/scala/com/gu/identity/paymentfailure/BrazeClient.scala
+++ b/payment-failure/src/main/scala/com/gu/identity/paymentfailure/BrazeClient.scala
@@ -31,3 +31,14 @@ class BrazeClient(config: Config) extends StrictLogging {
     }
   }
 }
+
+object BrazeClient {
+
+  // Valid keys of trigger properties that are utilised in Braze templates.
+  object TriggerProperties {
+    val emailToken = "emailToken"
+    val autoSignInToken = "autoSignInToken"
+    val abName = "abName"
+    val abVariant = "abVariant"
+  }
+}
--- a/payment-failure/src/main/scala/com/gu/identity/paymentfailure/BrazeEmailService.scala
+++ b/payment-failure/src/main/scala/com/gu/identity/paymentfailure/BrazeEmailService.scala
@@ -6,18 +6,36 @@ import com.gu.identity.paymentfailure.EncryptedEmailTest.Variant
 import com.gu.identity.paymentfailure.IdentityClient.{AutoSignInLinkRequestBody, IdentityEmailTokenRequest}
 import com.typesafe.scalalogging.StrictLogging
 
-class SendEmailService(
+// Make this an abstract trait to allow different implementations of sending an email.
+// See implementations for more context.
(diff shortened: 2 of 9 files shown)


payment-failure/src/main/scala/com/gu/identity/paymentfailure/BrazeClient.scala
```diff
@@ -31,3 +31,14 @@ class BrazeClient(config: Config) extends StrictLogging {
     }
   }
 }
+
+object BrazeClient {
+
+  // Valid keys of trigger properties that are utilised in Braze templates.
+  object TriggerProperties {
```
Mullefa, line 38: Hard code trigger properties to avoid typos etc.

payment-failure/src/main/scala/com/gu/identity/paymentfailure/BrazeEmailService.scala
```diff
@@ -0,0 +1,116 @@
...
+
+import BrazeClient.TriggerProperties
+import cats.syntax.either._
+import com.gu.identity.paymentfailure.EncryptedEmailTest.Variant
+import com.gu.identity.paymentfailure.IdentityClient.{AutoSignInLinkRequestBody, IdentityEmailTokenRequest}
+import com.typesafe.scalalogging.StrictLogging
+
+// Make this an abstract trait to allow different implementations of sending an email.
+// See implementations for more context.
+trait BrazeEmailService {
```
Mullefa, line 11: This is the trait discussed in the PR body.

payment-failure/src/main/scala/com/gu/identity/paymentfailure/LambdaService.scala
```diff
@@ -33,11 +33,20 @@ class LambdaService(sqsService: SqsService, sendEmailService: SendEmailService)
...
+    val sendEmailService = new DefaultBrazeEmailService(identityClient, brazeClient, config)
+    new LambdaService(sqsService, sendEmailService)
+  }
+
+  def encryptedEmailTest(config: Config): LambdaService = {
+    val identityClient = new IdentityClient(config)
+    val sqsService = new SqsService(config)
+    val brazeClient = new BrazeClient(config)
+    val encryptedEmailTest = new EncryptedEmailTest(identityClient)
+    val sendEmailService = new BrazeEmailServiceWithEncryptedBrazeEmailTest(brazeClient, encryptedEmailTest, config)
```
Mullefa, line 49: In future, if we want to change the change the `BrazeEmailService`, just create a new implementation and inject it into the `LambdaService`.

laurenemms (approved):

PR #34, opened 2019-05-03
Release AB test for Auto Sign In Tokens

laurenemms:
This releases the AB test for auto sign in tokens.

The test will involve around 20% of users, with 50% of these users receiving the auto sign in tokens and 50% in the control group. 

I tested this by increasing the range of the test so that my user id will be within the variant range and sending myself a payment failure email. I was taken directly to the 'update payment details' page. When my user id was not within range, I was to sign in with the 'journey' parameter set to 'paymentFailure' as expected.


payment-failure/src/main/scala/com/gu/identity/paymentfailure/BrazeEmailService.scala
```diff
@@ -90,13 +90,15 @@ class BrazeEmailServiceWithAbTest(
...
+    (for {
       variant <- variantGenerator.generateVariant(emailData.externalId, emailData.emailAddress)
       customFields = variantToCustomFields(variant)
       response <- sendEmailWithCustomFields(emailData, customFields)
     } yield {
-      logger.info(s"braze email sent with encrypted email test data - variant data: $variant")
+      logger.info(s"braze email sent with variant data for test - variant data: $variant")
       response
+    }).recoverWith {
+      case _: UserNotInVariantRange => sendEmailWithCustomFields(emailData, customFields = Map.empty)
```
laurenemms: If the error returned in the for comprehension is a result of the user not being in the variant range, still send them the email but without the auto sign in token. If the error is not a UserNotInVariantRange error, the email will not be sent.

Mullefa: Great! One suggestion: maybe amend the log statement at the start of this method: "attempting to send..." and if we get to this point in the code log an info message e.g. "user not in test range, sending regular email". Note that 'cos you've had `UserNotInVariantRange` extend a `RuntimeException` you can pass it as a `Throwable` to the second argument of `logger.info()`.

Mullefa: Another suggestion regarding logging, which was introduced by me (not you), but that I think makes sense to resolve before this test is released: to not include the auto sign-in token in any log entries. At the moment it will be on line 98 via the string interpolation of `$variant`. To address this you could change the `toString` implementation of `Variant` e.g. not include the `metadata` field. Alternatively, you could use the [`Show`](https://typelevel.org/cats/typeclasses/show.html) type class provided by cats.

Mullefa (commented):
Couple of suggestions on logging.

payment-failure/src/main/scala/com/gu/identity/paymentfailure/BrazeEmailService.scala
laurenemms: Have made these changes, thank you. Here is a screenshot from the logs for where the variant is logged now. 
Screen Shot 2019-05-03 at 15 15 09

Mullefa (approved):

PR #36, opened 2019-05-07
Changes to error handling 

laurenemms:
We reverted the Auto Sign In Token AB test because there was no error handling for when an Auto Sign In Token could not be generated. A cause of this is an 'Invalid User'. In a case like this, although we don't want to send the user an auto sign in token, we still want to send the email. For that reason, we still want to send the email.

This introduces some error handling to decode the JSON response and process the error. 

NOTE: I'm at a conference tomorrow. I know this still needs some more work, refactoring and testing but thought I'd put up a PR so that I can get any initial comments and have this merged sooner.


payment-failure/src/main/scala/com/gu/identity/paymentfailure/BrazeEmailService.scala
```diff
@@ -97,6 +97,11 @@ class BrazeEmailServiceWithAbTest(
     } yield {
       logger.info(s"braze email sent with encrypted email test data - variant data: $variant")
       response
+    }).recoverWith {
+      case err: UserIneligibleForAbTest => {
+        logger.info("user not in test range, sending regular email", err)
```
Mullefa: info message can be more generic: "user not eligible for test..."

payment-failure/src/main/scala/com/gu/identity/paymentfailure/IdentityClient.scala
```diff
@@ -23,17 +23,20 @@ class IdentityClient(config: Config) extends StrictLogging {
       .header("content-type", "application/json")
 
     Either.catchNonFatal(request.asString)
-      .leftMap[Throwable](err => networkError(path, err))
+      .leftMap(err => IdentityClientError.fromThrowable(err))
       .flatMap {
-        case res if res.is2xx => decode[Res](res.body).leftMap(err => decodingError(path, err))
-        case res => Left(non2xxError(path, res.code, res.body))
+        case res if res.is2xx => decode[Res](res.body).leftMap(err => IdentityClientError.fromCirceError(res.body, err))
```
Mullefa: Nice!

Some utility functions make this slightly clearer:
```scala
def decodeJson[A : Decoder](body: String): Either[IdentityClientError, A] =
  decode[A](body).leftMap(err => IdentityClientError.fromCirceError(body, err))

def decodeApiError(body: String): IdentityClientError =
  decodeJson[ApiError](body).merge
```
and then:
```scala
Either.catchNonFatal(request.asString)
  .leftMap(err => IdentityClientError.fromThrowable(err))
  .flatMap {
    case res if res.is2xx => decodeJson[Res](res.body)
    case res => Left(decodeApiError(res.body))
  }
```

payment-failure/src/main/scala/com/gu/identity/paymentfailure/IdentityClient.scala
```diff
@@ -47,18 +50,37 @@ object IdentityClient {
 
   @JsonCodec case class AutoSignInLinkResponseBody(token: String)
 
-  // Utility methods for creating exceptions related to the identity client.
-  // Since these are only getting called in the context of POST requests,
-  // can augment the messages with this information.
-  // Also, no point creating specific subtypes to model these errors at the moment,
-  // since the lambda takes the same action, regardless of the error.
+  @JsonCodec case class ApiError(errors: List[ApiError.Single]) extends IdentityClientError {
+    def invalidUser: Boolean = errors.exists(_.invalidUser)
```
Mullefa: `isInvalidUser` more idiomatic name

payment-failure/src/main/scala/com/gu/identity/paymentfailure/IdentityClient.scala
```diff
@@ -47,18 +50,37 @@ object IdentityClient {
...
+  }
+
+  object ApiError {
+    @JsonCodec case class Single(message: String) {
+      def invalidUser: Boolean = message == "Invalid user"
```
Mullefa: `isInvalidUser` more idiomatic name

payment-failure/src/main/scala/com/gu/identity/paymentfailure/IdentityClient.scala
```diff
@@ -47,18 +50,37 @@ object IdentityClient {
...
+    }
+  }
+
+  object IdentityClientError {
+    def fromThrowable(err: Throwable): IdentityClientError = UnderlyingError(err)
+    def fromCirceError(json: String, err: CirceError): IdentityClientError = UnknownJsonFormat(json, err)
+    def fromApiError(err: ApiError): IdentityClientError = err
+  }
 
-  def networkError(path: String, cause: Throwable): Exception =
```
Mullefa, line 56: double white space 🤣

payment-failure/src/main/scala/com/gu/identity/paymentfailure/IdentityClient.scala
```diff
@@ -47,18 +50,37 @@ object IdentityClient {
 
   @JsonCodec case class AutoSignInLinkResponseBody(token: String)
 
-  // Utility methods for creating exceptions related to the identity client.
-  // Since these are only getting called in the context of POST requests,
-  // can augment the messages with this information.
-  // Also, no point creating specific subtypes to model these errors at the moment,
-  // since the lambda takes the same action, regardless of the error.
+  @JsonCodec case class ApiError(errors: List[ApiError.Single]) extends IdentityClientError {
```
Mullefa: this is my bad (since it was ported from payment-api) but could this be defined under the sealed trait (and its companion object); reads a bit clearer that way e.g. with the rest of the classes that extend the base one

payment-failure/src/main/scala/com/gu/identity/paymentfailure/abtest/AutoSignInTest.scala
```diff
@@ -17,7 +17,7 @@ class AutoSignInTest(identityClient: IdentityClient) extends VariantGenerator {
           val body = AutoSignInLinkRequestBody(identityId, email)
           identityClient.createAutoSignInToken(body)
             .bimap(
-              err => new RuntimeException("unable to create auto sign-in token", err),
+              err => UserIneligibleForAbTest("unable to create auto sign-in token", Some(err)),
```
Mullefa: To send an email, the `BrazeEmailServiceWithAbTest`:
1. generates a variant
2. requests an email be sent including the variant data as meta data

It will only recover from a failure to execute these steps if the `UserIneligibleForAbTest`, in which case it will send an email without any variant meta data; otherwise the error will be propagated and the lambda will be executed again (i.e. retry to send the email with variant data). 

This behaviour means that the variant generator should only return an `UserIneligibleForAbTest` if we explicitly don't want to try and send the email again with the variant meta data, for example, there was a failure from the `IdentityApiClient` which is not likely to change over consecutive invocations of the lambda. 

Conversely, if we think that the error from the `IdentityApiClient` was ephemeral, we want to try and execute the lambda again, so don't return an `UserIneligibleForAbTest`.

Therefore, in the context of this generator I would recommend something like:
```scala
def handleIdentityClientError(err: IdentityClientError): Throwable =
  err match {
    // If the user is invalid for an auto sign-in token, 
    // return an UserIneligibleForAbTest,
    // so that the BrazeEmailServiceWithAbTest will send them a regular email
    case apiError: ApiError if apiError.invalidUser => UserIneligibleForAbTest("user not eligible for auto sign-in token", Some(apiError))
    // Otherwise return a RuntimeException
    // The BrazeEmailServiceWithAbTest will propagate this error,
    // meaning the lambda will be invoked again 
    // i.e. another attempt will be made to send the email to the reader
    // with the variant meta data
    case _ => new RuntimeException("unable to create auto sign-in token", err)
  }

override def generateVariant(identityId: String, email: String): Either[Throwable, Variant] =
  VariantGenerator.getSegmentId(identityId, from = 0, to = 0.2)
    .flatMap {
      case id if id < 0.1 => Right(controlVariant)
      case _ =>
        val body = AutoSignInLinkRequestBody(identityId, email)
        identityClient.createAutoSignInToken(body)
          .bimap(
            handleIdentityClientError,
            response => autoSignInTokenVariant(response.token)
          )
    }
```

Mullefa: A couple of remarks regarding this:
1. We could be more specific / declarative about the left hand side of the method `VariantGenerator.generateVariant()` e.g. `VariantGeneratorError` where:

    ```scala
    sealed trait VariantGeneratorError
    case class SendEmailWithoutVariantData(...) extends VariantGeneratorError
    case class RetrySendEmailWithVariantData(...) extends VariantGeneratorError
    ```
2. Or, we could encode the difference between sending without variant data or retrying with variant data by changing the return type of `VariantGenerator.generateVariant()`:

    ```scala
    def generateVariant(): Either[Throwable, Option[Variant]]
    ```
    where a `None` means send email without variant data, `Throwable` retry. In general though, I am weary of stacking monads as it leads to monad transformers or nested higher level method calls e.g. `.flatMap(_.flatMap(...))`

Ultimately, it's about making a decision on what we think is clearest. IMO I _think_ it's ok for now (maybe an additional comment on the `BrazeEmailServiceWithAbTest` to say _if the variant generator returns an error of type `UserIneligibleForAbTest` then..._)

payment-failure/src/main/scala/com/gu/identity/paymentfailure/abtest/EncryptedEmailTest.scala
```diff
@@ -18,7 +18,7 @@ class EncryptedEmailTest(identityClient: IdentityClient) extends VariantGenerato
       val requestBody = IdentityEmailTokenRequest(email)
       identityClient.encryptEmail(requestBody)
         .bimap(
-          err => new RuntimeException("unable to create email token", err),
+          err => UserIneligibleForAbTest("unable to create email token", Some(err)),
```
Mullefa: Same comments as on the `AutoSignInTest` apply. But I wouldn't bother updating this is test in this PR (in fact we could probably get rid of it in another PR).

Mullefa (commented):
A few comments. Let me know what you think.

payment-failure/src/main/scala/com/gu/identity/paymentfailure/IdentityClient.scala
laurenemms: Oh nice, thanks. I haven't seen merge before. So is it just when both cases are of the same type and you just return the value?

payment-failure/src/main/scala/com/gu/identity/paymentfailure/IdentityClient.scala
laurenemms: Oh no sorry I did actually move it so I could see what I was doing when rewriting executePostRequest and forgot to move it back. Have done now

payment-failure/src/main/scala/com/gu/identity/paymentfailure/abtest/EncryptedEmailTest.scala
laurenemms: I've added the same logic as Auto Sign In Test for now just for consistency and will make a PR to remove the email test after this has been merged

payment-failure/src/main/scala/com/gu/identity/paymentfailure/abtest/AutoSignInTest.scala
laurenemms: Just been through the tests again and see what you mean. I've made the changes you mentioned and added a comment in BrazeEmailServiceWithAbTest for now regarding when we are sending emails or not. I started looking into point 1. (having VariantGeneratorError trait) but I think it's quite clear now with the UserIneligibleForAbTest case class and comments what's happening. What do you think?

payment-failure/src/main/scala/com/gu/identity/paymentfailure/IdentityClient.scala
Mullefa: Yeah, or you could merge to a super type as in [this](https://www.scala-lang.org/api/current/scala/util/Either$$MergeableEither.html) example where an `Either[List[Int], Vector[Int]]` is being merged.

payment-failure/src/main/scala/com/gu/identity/paymentfailure/abtest/AutoSignInTest.scala
Mullefa: Yeah, I think it's fine as is. Interesting to think about the alternatives though. At least for me 😂🙈

Mullefa (approved):

PR #39, opened 2019-05-20
handle user not found error

Mullefa:
Failing to handle this caused the dead-letter queue alert. If a user isn't found, send a 'regular' email. See inline comment for justification regarding this error handling.


dominickendrick (approved):
👍

PR #40, opened 2019-05-20
remove alarm which alerts on lambda error

Mullefa:
As discussed in pager duty email thread, don't alert on lambda error, since if max retries are exceeded, we'll get an alert on the dead letter queue anyway i.e. only alert on error if maximum number of retries have been exceeded.


jfsoul (approved):
As long as you're confident the dead letter queue alarm is working as intended 👍

PR #41, opened 2019-06-13
Formstack Consent Lambda

laurenemms:
There is currently a python script running on a server daily at 10.30am that grabs all Formstack consent submissions and sends the email and consent name to Identity so that we can send users confirmation emails. When they click the link in these emails, we create an Identity account for them and they will be subscribed to that newsletter. 

It was raised to us by the Data Insights team that there was a large difference in the number of users who sign up to receive these newsletters and the number of users who have identity accounts. A small difference is expected as:
1. There is not 100% conversion rate of users who will click the link in the confirmation email (creating an identity account)
2. The conversion rate is made even lower as users will only receive this confirmation email when the script runs, rather than when they fill out the form

However, the difference highlighted by DI was much larger than expected, triggering an investigation into the cause. It was found that the API key used for Formstack had expired and I was given a new one, which was then added to the script. The script successfully ran the following day (sending over 7000 emails as it had not run for 8 weeks).

In order to ensure that we can keep a closer eye on this script in the future, I have created migrated the script to a lambda, which will live in Identity's AWS account.

I will remove DO NOT MERGE label once the following have been added:
- Tests
- Cloudformation and parameter config set up
- Logging

### DIFF
--- a/formstack-consents/src/main/scala/com/gu/identity/formstackconsents/FormstackClient.scala
+++ b/formstack-consents/src/main/scala/com/gu/identity/formstackconsents/FormstackClient.scala
@@ -56,9 +56,7 @@ class FormstackClient(config: DevConfig) extends StrictLogging {
 }
 
 object FormstackClient {
-  implicit val config: Configuration = Configuration.default
-  // Scala won't allow 'type' as an argument so specifying it used @JsonKey
-  @ConfiguredJsonCodec case class FormstackConsent(field: String, label: String, @JsonKey("type") consentType: String, value: String)
+  @JsonCodec case class FormstackConsent(field: String, label: String, `type`: String, value: String)
   @JsonCodec case class FormstackSubmission(data: Map[String, FormstackConsent])
   @JsonCodec case class FormstackResponse(submissions: List[FormstackSubmission], pages: Int)
 }
--- a/formstack-consents/src/main/scala/com/gu/identity/formstackconsents/FormstackClient.scala
+++ b/formstack-consents/src/main/scala/com/gu/identity/formstackconsents/FormstackClient.scala
@@ -1,0 +1,65 @@
+package com.gu.identity.formstackconsents
+
+import com.gu.identity.globalConfig.DevConfig
+import cats.syntax.either._
+import com.gu.identity.formstackconsents.FormstackClient.FormstackResponse
+import com.typesafe.scalalogging.StrictLogging
+import io.circe.parser.decode
+import scalaj.http.Http
+import io.circe.generic.JsonCodec
+import io.circe.generic.extras.{Configuration, ConfiguredJsonCodec, JsonKey}
+import org.joda.time.DateTime
+
+class FormstackClient(config: DevConfig) extends StrictLogging {
+
+  def requestForMultiplePages(res: FormstackResponse, newsletter: Newsletter): List[FormstackResponse] = {
+    val pagesList = (2 to res.pages).toList
+    pagesList.foldLeft(List(res)){ (acc, page) => {
+      getSubmissionsForGivenPage(newsletter.formId, page) match {
+        case Left(_) => List(res)
+        case Right(response) => response :: acc
+      }
+    }}
+  }
+
+  def getSubmissionsForGivenPage(newsletterId: String, page: Int): Either[Throwable, FormstackResponse] = {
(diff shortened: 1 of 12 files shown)


formstack-consents/src/main/scala/com/gu/identity/formstackconsents/IdentityClient.scala
```diff
@@ -0,0 +1,49 @@
...
+    //  }
+
+    implicit val circeConfig: Configuration = Configuration.default.copy(
+      transformMemberNames = {
+        case "listType" => newsletter.listType
+        case other => other
+      }
+    )
+
+    @ConfiguredJsonCodec case class IdentityRequest(email: String, listType: List[String])
```
laurenemms, line 28: The reason I have defined this case class (and above config) within the class and not in a companion object is because a key of the class (listType) changes depending on the newsletter (see comment starting on line 14)

formstack-consents/src/main/scala/com/gu/identity/formstackconsents/Lambda.scala
```diff
@@ -0,0 +1,19 @@
+package com.gu.identity.formstackconsents
+
+import com.gu.identity.globalConfig.DevConfig
+
+object Lambda extends App {
+
+  val newsletters: List[Newsletter] = List(Holidays, Students, Universities, Teachers, Masterclasses, SocietyWeekly, EdinburghFestivalDataCollection)
+
+  def handler(): Unit = {
+    // TODO: look into integrating parameter store through cloudformation
```
laurenemms, line 10: Config is currently set within the project but once I look into cloudformation and parameter store, I will change this.

formstack-consents/src/main/scala/com/gu/identity/formstackconsents/LambdaService.scala
```diff
@@ -0,0 +1,33 @@
...
+        logger.error(errorMessage)
+        Left(new Throwable(errorMessage))
+    }
+  }
+
+  def sendConsentToIdentity(consent: FormstackConsent, newsletter: Newsletter): Either[Throwable, Unit] = {
+    identityClient.sendConsentToIdentity(consent, newsletter)
+  }
+
+  def getConsentsAndSendToIdentity(newsletter: Newsletter): Either[Throwable, Unit] = {
```
laurenemms, line 26: If anyone can help find a better way to write this method I'd appreciate any comments!

lmath:
I'm so glad this code is in a place we can find 🎉

laurenemms:
> I'm so glad this code is in a place we can find 🎉

formstack-consents/src/main/scala/com/gu/identity/formstackconsents/FormstackClient.scala
```diff
@@ -0,0 +1,65 @@
...
+    request.flatMap(res => res.pages match {
+      case 1 => Right(List(res))
+      case _ => Right(requestForMultiplePages(res, newsletter))
+    })
+  }
+}
+
+object FormstackClient {
+  implicit val config: Configuration = Configuration.default
+  // Scala won't allow 'type' as an argument so specifying it used @JsonKey
```
dominickendrick: I think if you put it in backticks `` you can use `type`

laurenemms: You're right - thank you

Mullefa:
Haven't had a chance to look through the code yet, but just want to say that it's so appreciated providing (such detailed) context in the PR.

formstack-consents/src/main/scala/com/gu/identity/formstackconsents/LambdaService.scala
dominickendrick: After chatting we will try something like 
```
def processPageResponses(pageResponses: List[FormstackClient.FormstackResponse], newsletter: Newsletter): List[Either[Throwable,HttpResponse[String]]] = {
    val submissions: List[FormstackClient.FormstackSubmission] = pageResponses.flatMap(_.submissions)
    val emails: List[Either[Throwable, FormstackConsent]] = submissions.map(getEmailConsentFromSubmission)
    emails.map{ consentsOrErrors =>
      consentsOrErrors.flatMap(consent => identityClient.sendConsentToIdentity(consent, newsletter))
    }
  }
```

PR #42, opened 2019-06-24
Formstack Consents

laurenemms:
There is currently a python script running on a server daily at 10.30am that grabs all Formstack consent submissions and sends the email and consent name to Identity so that we can send users confirmation emails. When they click the link in these emails, we create an Identity account for them and they will be subscribed to that newsletter.

It was raised to us by the Data Insights team that there was a large difference in the number of users who sign up to receive these newsletters and the number of users who have identity accounts. A small difference is expected as:

There is not 100% conversion rate of users who will click the link in the confirmation email (creating an identity account)
The conversion rate is made even lower as users will only receive this confirmation email when the script runs, rather than when they fill out the form
However, the difference highlighted by DI was much larger than expected, triggering an investigation into the cause. It was found that the API key used for Formstack had expired and I was given a new one, which was then added to the script. The script successfully ran the following day (sending over 7000 emails as it had not run for 8 weeks).

In order to ensure that we can keep a closer eye on this script in the future, I have created migrated the script to a lambda, which will live in Identity's AWS account.

This operates slightly differently to the old script in a number of ways: 
- No more calls to Formstack 
- Only one consent is processed each time
- Each form will be set up with a webhook so that the lambda is triggered _each time_ a consent is submitted. This means that a user will now receive a confirmation email as soon as they submit the form

TODO:
- Added webhooks to forms via Formstack
- Stop the current Formstack Consents script


formstack-consents/README.md
```diff
@@ -0,0 +1,5 @@
+# Formstack Consents Lambda
+
+This lambda is responsible for making calls to Formstack, getting the email address of each user who has signed up to newsletters via Formstack and making a subsequent POST request to Identity API /consent-email with the user's email address and the newsletter they signed up to. This will then trigger a confirmation email for that user.
```
Mullefa: Out of date.

formstack-consents/cloud-formation.yaml
```diff
@@ -0,0 +1,106 @@
...
+          Action:
+          - logs:CreateLogGroup
+          - logs:CreateLogStream
+          - logs:PutLogEvents
+          Resource: !Sub arn:aws:logs:${AWS::Region}:${AWS::AccountId}:log-group:/aws/lambda/FormstackConsentsLambda-${Stage}:log-stream:*
+      - Version: '2012-10-17'
+        Statement:
+        - Effect: Allow
+          Action:
+          - ssm:GetParameter
```
Mullefa: Is this needed?

formstack-consents/cloud-formation.yaml
```diff
@@ -0,0 +1,106 @@
...
+          Resource: !Sub arn:aws:ssm:${AWS::Region}:${AWS::AccountId}:parameter/FormstackConsentsLambda-${Stage}
+    Events:
+      GetApi:
```
Mullefa: Shouldn't this be called something like `FormstackConsentApi` ?

formstack-consents/cloud-formation.yaml
```diff
@@ -0,0 +1,106 @@
...
+        Type: Api
+        Properties:
+          RestApiId:
+            Ref: ServlessRestApi
+          Path: /consent
+          Method: POST
+
+  ServerlessRestApi:
```
Mullefa, line 51: From the [AWS::Serverless::Api documentation](https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#awsserverlessapi), I don't think this needs to be explicitly created. Do you know what the benefits are of doing so?

formstack-consents/cloud-formation.yaml
```diff
@@ -0,0 +1,106 @@
...
+      FunctionName: !Ref FormstackConsentsLambda
+      Principal: apigateway.amazonaws.com
+
+  FormstackConsentsLambdaLogGroup:
+    Type: AWS::Logs::LogGroup
+    Properties:
+      LogGroupName: !Sub /aws/lambda/${FormstackConsentsLambda}
+      RetentionInDays: 14
+
+  FormstackConsentsDeadLetterQueue:
```
Mullefa: This resource isn't currently used.

formstack-consents/cloud-formation.yaml
```diff
@@ -0,0 +1,106 @@
...
+    Type: AWS::Lambda::Permission
+    DependsOn:
+      - FormstackConsentsLambda
+      - ServerlessRestApi
+    Properties:
+      Action: lambda:InvokeFunction
+      FunctionName: !Ref FormstackConsentsLambda
+      Principal: apigateway.amazonaws.com
+
+  FormstackConsentsLambdaLogGroup:
```
Mullefa: Don't think this is needed.

formstack-consents/cloud-formation.yaml
```diff
@@ -0,0 +1,106 @@
...
+            post:
+              x-amazon-apigateway-integration:
+                # The HTTP method used in the integration request. For Lambda function invocations, the value must be POST.
+                httpMethod: POST
+                type: aws_proxy
+                uri:
+                  Fn::Sub: arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${FormstackConsentsLambda.Arn}/invocations
+              responses: {}
+
+  FormstackConsentsLambdaPerms:
```
Mullefa, line 72: Don't think this is needed.

formstack-consents/cloud-formation.yaml
```diff
@@ -0,0 +1,106 @@
...
+          idapiHost: !Sub ${IdentityApiEndpoint}
+          idapiAccessToken: !Sub ${IdentityAPIKey}
+      Handler: com.gu.identity.formstackconsents.Lambda::handler
+      Runtime: java8
+      CodeUri:
+        Bucket: identity-lambda
+        Key: !Sub identity/${Stage}/formstack-consents-lambda/main.jar
+      Policies:
+      - Version: '2012-10-17'
+        Statement:
```
Mullefa: Don't think this log policy is needed explicitly.

formstack-consents/src/main/resources/logback.xml
```diff
@@ -0,0 +1,13 @@
+<configuration>
+    <contextName>formstack-consents</contextName>
+
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
```
Mullefa: With this appender, are multiline log statements considered as individual logs in cloudwatch? If so you can use the [`AwsLambdaAppender`](https://github.com/guardian/identity-processes/blob/master/payment-failure/src/main/resources/logback.xml#L3).

formstack-consents/src/main/scala/com/gu/identity/formstackconsents/IdentityClient.scala
```diff
@@ -0,0 +1,66 @@
...
+
+import io.circe.syntax._
+import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyResponseEvent
+import com.gu.identity.formstackconsents.Lambda.{FormstackSubmission, Config}
+import com.typesafe.scalalogging.StrictLogging
+import io.circe.generic.extras._
+import scalaj.http.{Http, HttpResponse}
+import scala.util.Try
+
+class IdentityClient(config: Config) extends StrictLogging {
```
Mullefa, line 15: It's not really an issue for this lambda, but just a note that this class is doing quite a few things:

- interacting with the identity API
- understanding a Formstack consent and how to transform one to an identity request
- transforming a response from the identity API to an `APIGatewayProxyResponseEvent`
- understanding (statically) what news letters are supported

As an example of why this might be an issue in the context of a larger application: you wouldn't be able to use this class to post a consent from a different source.

My preference would be to break it down into a pure identity API client and a Formstack service, however, it's pretty subjective, so can leave as is if you want.

formstack-consents/src/main/scala/com/gu/identity/formstackconsents/IdentityClient.scala
```diff
@@ -0,0 +1,66 @@
...
+
+  val newsletters: List[Newsletter] = List(Holidays, Students, Universities, Teachers, Masterclasses, SocietyWeekly, EdinburghFestivalDataCollection)
+
+  def getResponseFromIdentity(formstackSubmission: FormstackSubmission, newsletter: Newsletter): Option[HttpResponse[String]] = {
```
Mullefa: method name like `updateConsent()` ?

formstack-consents/src/main/scala/com/gu/identity/formstackconsents/IdentityClient.scala
```diff
@@ -0,0 +1,66 @@
...
+    // The key in the JSON sent to Identity depends on the listType. Sometimes the listType is 'set-lists' with a value of the consent name,
+    // and sometimes the listType is 'set-consents'. See example below.
+    //  {
+    //    "email" : "lauren.emms@guardian.co.uk",
+    //    "set-consents" : "holidays"
+    //  }
+
+    implicit val circeConfig: Configuration = Configuration.default.copy(
```
Mullefa: Interesting approach. Ideally, you don't want to tie down the `IdentityRequest` class to a specific class instance, however, I can see why it's necessary if you're using the `@ConfiguredJsonCodec` annotation. Another approach could be to define `IdentityRequest` statically somewhere (e.g. the companion object) along with a function: `Newsletter => Encoder[IdentityRequest]` which could then be called within this method to define an implicit encoder for the given news letter.

formstack-consents/src/main/scala/com/gu/identity/formstackconsents/IdentityClient.scala
```diff
@@ -0,0 +1,66 @@
...
+
+    Try {
+      Http(s"${config.idapiHost}/consent-email")
+        .headers(("Authorization", config.idapiAccessToken), ("Content-type", "application/json"), ("Accept", "text/plain"))
+        .postData(requestBody.asJson.noSpaces)
+        .asString
+    }.toOption
+  }
+
+  def handleResponseFromIdentity(response: Option[HttpResponse[String]], formstackSubmission: FormstackSubmission, newsletter: Newsletter): Option[APIGatewayProxyResponseEvent] = {
```
Mullefa, line 29: Don't think the return type should be an `Option` - if the post to the identity API fails, we can encode this information using the `APIGatewayProxyResponseEvent`

formstack-consents/src/main/scala/com/gu/identity/formstackconsents/Lambda.scala
```diff
@@ -0,0 +1,52 @@
...
+  def getConfig: Option[Config] = {
+    for {
+      idapiHost <- getEnvironmentVariable("idapiHost")
+      idapiAccessToken <- getEnvironmentVariable("idapiAccessToken")
+    } yield Config(idapiHost, idapiAccessToken)
+  }
+
+  implicit val circeConfig: Configuration = Configuration.default
+  // this will change depending on form
+  @ConfiguredJsonCodec case class FormstackSubmission(@JsonKey("FormID") formId: String, @JsonKey("email_address") emailAddress: String)
```
Mullefa: More idiomatic for this to be defined in it's own class.

formstack-consents/src/main/scala/com/gu/identity/formstackconsents/IdentityClient.scala
Mullefa: Or you could keep it as `Option` which would be defined iff the `Option[HttpResponse[String]]` is defined (if the response is non 2xx, then you could use its status to set the status of the `APIGatewayProxyResponseEvent`)

formstack-consents/src/main/scala/com/gu/identity/formstackconsents/Lambda.scala
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  def handler(event: APIGatewayProxyRequestEvent): APIGatewayProxyResponseEvent = {
+    logger.info("event" + event.toString)
+    logger.info("body" + event.getBody)
+//    Uncomment when ready to test
+    (for {
+      config <- getConfig
+      identityClient = new IdentityClient(config)
+      formstackSubmission <- decodeFormstackSubmission(event.getBody)
```
Mullefa: Need to check that the Formstack consent has been sent from Formstack; for example, using a [shared secret](https://developers.formstack.com/docs/webhook-setup#section-shared-secret).

Mullefa (commented):
Only thing that blocks this PR is to check that the consent has been sent from Formstack; other points are relatively minor.

Couple more general points:
- Re cloudformation, suggestions made with not complete confidence (haven't used SAM before), but I don't think as much of it is required explicitly e.g. here is a [relatively new example](https://github.com/cpliakas/aws-sam-golang-example/blob/master/template.yaml) which forgoes some of the sections you have added.
- `Option` instead of `Either` ? 😔😅
- Would be awesome to create a utility script (using `sam local start-api` ) to facilitate testing changes locally.

formstack-consents/src/main/resources/logback.xml
laurenemms: Thank you

formstack-consents/src/main/scala/com/gu/identity/formstackconsents/IdentityClient.scala
laurenemms: I have created a companion object with a method in, it needs the implicit val set within the same scope as defining the case class. Is this what you meant by the Newsletter => Encoder[IdentityRequest] function?

formstack-consents/src/main/scala/com/gu/identity/formstackconsents/IdentityClient.scala
laurenemms: Thank you, have kept this as an option and used the request code as you suggested.

Mullefa (approved):
Looks great 💯; as you mentioned would be good to have an alarm in a subsequent PR to alert us on errors. 

Feel free to merge and deploy to PROD!

PR #43, opened 2019-06-26
move to default Braze service

Mullefa:
There was an upstream change in membership workflow which has changed the type of external id. Reverting to the default service and deferring thinking through / addressing the implication of this till tomorrow.

Doing this will stop the auto sign-in AB test.

### DIFF
--- a/payment-failure/src/main/scala/com/gu/identity/paymentfailure/BrazeEmailService.scala
+++ b/payment-failure/src/main/scala/com/gu/identity/paymentfailure/BrazeEmailService.scala
@@ -2,7 +2,6 @@ package com.gu.identity.paymentfailure
 
 import BrazeClient.TriggerProperties
 import cats.syntax.either._
-import com.gu.identity.paymentfailure.IdentityClient.{AutoSignInLinkRequestBody, IdentityEmailTokenRequest}
 import com.gu.identity.paymentfailure.abtest._
 import com.typesafe.scalalogging.StrictLogging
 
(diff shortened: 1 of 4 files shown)


payment-failure/src/main/scala/com/gu/identity/paymentfailure/BrazeEmailService.scala
```diff
@@ -27,47 +26,10 @@ object BrazeEmailService {
...
-    identityClient.createAutoSignInToken(AutoSignInLinkRequestBody(identityId, email))
-      .fold(
-        err => {
-          // TODO: monitor these errors
-          logger.error(s"unable to create auto sign in token for identity id $identityId and email $email", err)
-          Option.empty[String]
-        },
-        response => Some(response.token)
-      )
+class DefaultBrazeEmailService(brazeClient: BrazeClient, config: Config) extends BrazeEmailService with StrictLogging {
```
Mullefa, line 29: Conceptually, I think it makes sense for the default service to not bother with any tokens that facilitate sign-in. It's also convenient in the context of this PR 😉

payment-failure/src/main/scala/com/gu/identity/paymentfailure/Lambda.scala
```diff
@@ -59,11 +59,7 @@ object Lambda extends StrictLogging {
       throw err
     }
 
-    // Currently running the auto sign in test
-    // TODO: switch back to using DefaultBrazeEmailService when auto sign in token test is finished
-    val identityClient = new IdentityClient(config)
-    val autoSignInTokenTest = new AutoSignInTest(identityClient)
-    val lambdaService = LambdaService.withAbTest(config, autoSignInTokenTest)
+    val lambdaService = LambdaService.default(config)
```
Mullefa, line 62: Switch to using default service instead of one that implements an AB test

lmath (approved):
🌟

PR #47, opened 2019-08-07
Add Opt In checks to formstack-consent lambda

vlbee:
This change is required by 9/10/19.

# What does this change? 

Adds a `checkHasOptedIn` check to the Lambda's IdentityClient.sendConsentToIdentity control flow to ensure  a consent email is triggered only in the appropriate cases.

### Supporting changes: 
- Adds `MarketingConsent` type for forms that specifically require opt in checks to trigger a consent email from IDAPI
- Updates `FormSubmission` case class to include `opt_in: Option[String]`

# Why is this needed?

Marketing also uses Formstack forms with the option to 'opt in' to additional marketing consents for their campaigns, in this case during a number of summer events.  They wanted to use our existing Professional Networks newsletter signup form integration to collect this marking consent the same day forms are submitted via this existing lambda infrastructure. 

Forms are created and managed by Marketing in Formstack, where a webhook to the lamba is triggered if some one opts in. 

However, as these Formstack settings are not exclusively under our control, with this PR we perform an additional check in the lambda itself to ensure we are never collecting consent and user data inappropriately.

### DIFF
--- a/formstack-consents/src/main/scala/com/gu/identity/formstackconsents/Lambda.scala
+++ b/formstack-consents/src/main/scala/com/gu/identity/formstackconsents/Lambda.scala
@@ -52,7 +52,12 @@ object Lambda extends StrictLogging {
   }
 }
 
-case class FormstackSubmission(formId: String, emailAddress: String, handshakeKey: String)
+case class FormstackSubmission(
+  formId: String,
+  emailAddress: String,
+  handshakeKey: String,
+  opt_in: Option[Boolean] = None // if required, request the formstack checkbox hidden label be "opt_in"
+)
 
 object FormstackSubmission {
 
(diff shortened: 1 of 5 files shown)


vlbee:
@Mullefa  As discussed, instead of now conditionally decoding the form submission, I have updated the `formstackSubmission` model to include an `opt-in: Option[String]`

This is then checked in the `IdentityClient.checkHasOptedIn` whether to send the consent request or not.

Mullefa (approved):
Nice 💯

formstack-consents/src/main/scala/com/gu/identity/formstackconsents/Lambda.scala
```diff
@@ -52,23 +52,43 @@ object Lambda extends StrictLogging {
...
     cursor.downField("email_address").as[String]
       .orElse(cursor.downField("your_email_address").as[String])
+        .orElse(cursor.downField("email").as[String])
+  }
+
+  // Form submissions of type MarketingConsent require the user to 'opt in' to consents, not just submit.
+  // While we can conditionally trigger lambda webhook within Formstack, these settings are not exclusively
+  // under our control so we perform an additional checks in code to ensure we are never collecting consent and user data inappropriately.
+  // Any forms with required opt ins should be added in Newsletter.scala. Request the formstack checkbox hidden label be "opt_in"
+  private def getConsentOptInField(cursor: HCursor): Option[Boolean] = {
```
mchv, line 74: @vlbee This is minor but you should use `/**` for multiline comment rather than `//`

formstack-consents/src/main/scala/com/gu/identity/formstackconsents/Newsletter.scala
```diff
@@ -42,8 +42,24 @@ case object SocietyWeekly extends Newsletter {
...
 case object EdinburghFestivalDataCollection extends Newsletter {
   val formId = "3163410"
   val listType = "set-consents"
   val consent = "supporter"
+}
+
+// Some marketing Forms have an opt in checkbox for consent collection.
+// Formstack form can be setup with conditional logic to trigger the lambda webhook,
+// but we want tp check again when the submission is decoded as a precaution.
+// Add new MarketingConsent to newsletters and optInForms values in IdentityClient
```
mchv, line 55: @vlbee idem

PR #51, opened 2020-03-31
Event brite consents sync lambda

QuarpT:
# Eventbrite consents sync Lambda

Scala lambda which hits the Eventbrite API and finds answers where users have ticked the 'hear more about guardian news and media' check box.

For positive answers the lambda hits the consent-emails endpoint idapi to trigger a confirmation email.

## Deployments

There is no CODE environment for Eventbrite. For testing we have a DEV Lambda which is configured with `isDebug=true`. This prevents the lambda
from updating the consents in Identity.

This Lambda was migrated from a defunct Python script running as a cron job.

### DIFF
--- a/eventbrite-consents/README.md
+++ b/eventbrite-consents/README.md
@@ -1,10 +1,45 @@
 # Eventbrite consents sync Lambda
 
-Scala lambda which hits the Eventbrite API and finds answers where users have ticked the 'hear more about guardian news and media' check box.
+The service is a Scala lambda which sends emails to users who tick the 'hear more about guardian news and media' check box in Evenbrite.
 
-For positive answers the lambda hits the consent-emails endpoint idapi to trigger a confirmation email.
+## Lambda Details
+
+Eventbrite is a third party service where users can sign up for events. As part of the Eventbrite flow users can click a checkbox labelled
+
+```
+Please tick if you are interested in hearing about these products and services from Guardian News & Media 
+```
+
+The Lambda retrieves a list of emails for users who responded positively to the question from the Eventbrite API. The lambda passes the emails to 
+the `/consent-emails` endpoint on Identity API which triggers confirmation emails for the events consent.
 
 ## Deployments
 
 There is no CODE environment for Eventbrite. For testing we have a DEV Lambda which is configured with `isDebug=true`. This prevents the lambda
-from updating the consents in Identity.
+from updating the consents in Identity. 
+
+## Running Locally
+
+To run the app locally set the environment variables for the app (including API tokens). The parameters used
+in DEV and PROD can be found in the Lambda environment configuration on the AWS console. Be careful, if `isDebug=false` then
+emails could potentially be sent to users if idapiHost and idapiAccessToken are set.
+
+
+Run with SBT
+```
+export idapiHost=not_used_in_debug
+export idapiAccessToken=not_used_in_debug
+export masterclassesToken=set_this
+export eventsToken=set_this
+export syncFrequencyHours=4
+export isDebug=true
+sbt run
+```
+
+or run the com.gu.identity.eventbriteconsents.Lambda.main function with environment variables set in Intellij.
+
+## Unit tests
+
+```
+sbt test
+```
--- a/eventbrite-consents/README.md
+++ b/eventbrite-consents/README.md
@@ -1,0 +1,10 @@
+# Eventbrite consents sync Lambda
+
+Scala lambda which hits the Eventbrite API and finds answers where users have ticked the 'hear more about guardian news and media' check box.
+
+For positive answers the lambda hits the consent-emails endpoint idapi to trigger a confirmation email.
+
+## Deployments
+
+There is no CODE environment for Eventbrite. For testing we have a DEV Lambda which is configured with `isDebug=true`. This prevents the lambda
+from updating the consents in Identity.
--- a/eventbrite-consents/cloud-formation.yaml
+++ b/eventbrite-consents/cloud-formation.yaml
@@ -30,6 +30,11 @@ Mappings:
       value: true
     PROD:
       value: false
+  SyncFrequencyMap:
+    values:
+      # syncFrequency needs to match cron
+      cronExpression: "cron(0 0,4,8,12,16,20 * * ? *)"
+      syncFrequencyHours: 4
 
 Resources:
   TopicPagerDutyAlerts:
@@ -53,8 +58,7 @@ Resources:
           idapiAccessToken: !Ref IdapiAccessToken
           masterclassesToken: !Ref MasterclassesToken
           eventsToken: !Ref EventsToken
-          # update the cron expression when updating this
-          syncFrequencyHours: 4
+          syncFrequencyHours: !FindInMap [SyncFrequencyMap, values, syncFrequencyHours]
           isDebug: !FindInMap [IsDebugMap, !Ref Stage, value]
       Handler: com.gu.identity.eventbriteconsents.Lambda::handler
       Runtime: java8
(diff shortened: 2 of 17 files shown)


eventbrite-consents/README.md
```diff
@@ -0,0 +1,10 @@
+# Eventbrite consents sync Lambda
```
rebecca-thompson, line 1: Should there be some more context around what we use Eventbrite for? + any other useful info re tests, running it locally, etc

eventbrite-consents/cloud-formation.yaml
```diff
@@ -0,0 +1,102 @@
...
+    Description: environment name
+    Type: String
+    AllowedValues:
+      - DEV
+      - PROD
+
+Conditions:
+  IsProd: !Equals
+    - !Ref Stage
+    - PROD
```
vlbee, line 25: Didn't realise you could do this in cfn yml and filter alerts. 👍

rebecca-thompson (approved):

eventbrite-consents/src/main/scala/com/gu/identity/eventbriteconsents/Lambda.scala
```diff
@@ -0,0 +1,25 @@
...
+
+
+object Lambda extends LazyLogging {
+  val config: LambdaConfig = LambdaConfig.loadFromEnvironment()
+  val eventbriteClient: EventbriteClient = new EventbriteClient
+  val idapiClient: IdentityClient = new IdentityClient(config.idapiHost, config.idapiAccessToken)
+  val consentsService: ConsentsService = new ConsentsService(config, eventbriteClient, idapiClient)
+
+  def handler(event: APIGatewayProxyRequestEvent): APIGatewayProxyResponseEvent = {
+    consentsService.syncConsents()
```
vlbee, line 18: Why do we need to call this in both handler and main?

eventbrite-consents/src/main/scala/com/gu/identity/eventbriteconsents/services/ConsentsService.scala
```diff
@@ -0,0 +1,49 @@
...
+import com.gu.identity.eventbriteconsents.clients.{EventbriteClient, IdentityClient}
+import com.gu.identity.eventbriteconsents.config.LambdaConfig
+import com.typesafe.scalalogging.LazyLogging
+
+class ConsentsService(config: LambdaConfig, eventbriteClient: EventbriteClient, identitiyClient: IdentityClient) extends LazyLogging {
+
+  def syncConsents(): Unit = {
+    val lastRun = Instant.now().minus(config.syncFrequencyHours, ChronoUnit.HOURS)
+    val emailAddresses = findConsentEmails(lastRun, config.masterclassesToken) ++ findConsentEmails(lastRun, config.eventsToken)
+    logger.info(s"Number of email address to sync ${emailAddresses.size}")
```
vlbee: Might be useful to include `lastRun` var in this log message?

eventbrite-consents/src/main/scala/com/gu/identity/eventbriteconsents/services/ConsentsService.scala
```diff
@@ -0,0 +1,49 @@
...
+    for {
+      email <- emailAddresses
+    } yield if (config.isDebug) {
+        logger.info(s"Debug flag is set, not syncing $email")
+    } else {
+      identitiyClient.updateEventConsent(email)
+      logger.info(s"Synced $email")
```
vlbee, line 22: If we are logging a successfully synced email here, should we not also log the email when `identitiyClient.updateEventConsent(email)` fails?  In case we need to replay any failed updates to Identity?

eventbrite-consents/src/main/scala/com/gu/identity/eventbriteconsents/clients/IdentityClient.scala
```diff
@@ -0,0 +1,21 @@
...
+      .headers(
+        "X-GU-ID-Client-Access-Token" -> s"Bearer $idapiAccessToken",
+        "Content-type" -> "application/json"
+      )
+      .postData(IdapiConsentUpdate(emailAddress, Vector("events")).asJson.noSpaces)
+      .asString
+
+    if (result.code != 200) throw new RuntimeException(s"Unexpected response from idapi ${result.code} ${result.body}")
+  }
+}
```
vlbee: It's unclear to me how we are handling errors on this end. If Identity is down, will this retry to update the email? Might we not loose the occasional consent here as each one is a separate request to Identity that has to succeed?

vlbee:
It's unclear to me how to develop this locally and pass in that isDebug var? Could you add some more instructions in the README? If I worked with lambdas more regularly, this might be obvious but at the moment it's not so even a link to aws docs would be helpful.

The Formstack consents lambda uses AWS SAM with instructions in that README so if this one is different I think even more important to flesh out the details. 

@QuarpT  I really enjoyed reading your code (compared to another repo I have recently been working on) and learning about [Vectors vs Lists](https://docs.scala-lang.org/overviews/collections/performance-characteristics.html) in Scala. Which type of operation were you optimizing performance for with Vector here? Is it the loop in `findConsentEmails` that will be faster vs using a List?

eventbrite-consents/README.md
QuarpT: 👍 I will add some more info

eventbrite-consents/src/main/scala/com/gu/identity/eventbriteconsents/Lambda.scala
QuarpT: main is for testing locally. I'll add something to the readme

eventbrite-consents/src/main/scala/com/gu/identity/eventbriteconsents/clients/IdentityClient.scala
QuarpT: I'm concerned about implementing a retry mechanism since sending emails is not idempotent - retrying the action might result in someone being spammed in the scenario where identity successfully sends the email but some error (e.g. network failure) prevents the successful result being returned to the lambda. 

The choice is:

- users always receive at least one or more emails when they consent (where they are potentially spammed by identity)
- a user receives no more than one email (the user might not receive an email on failure)

There are pager duty alerts for lambda failures. I will add more logging to make it clear which emails the service received successful responses for, and which ones were never triggered.

eventbrite-consents/src/main/scala/com/gu/identity/eventbriteconsents/services/ConsentsService.scala
QuarpT: 👍 I will add more logging to make it clear which emails the service received successful responses for, which one failed and which ones were never triggered.

eventbrite-consents/src/main/scala/com/gu/identity/eventbriteconsents/clients/IdentityClient.scala
vlbee: ok. This makes sense as it's just the email getting triggered not the actual consent being added to Identity. I guess we need to accept that we will loose the occasional signup but I imagine this will be rare. If we set up an alert for a (highish) threshold of failures we will a least know if all the emails in a sync failed because IDAPI was having issues at that moment.

vlbee (approved):
✨

QuarpT:
> It's unclear to me how to develop this locally and pass in that isDebug var? Could you add some more instructions in the README? If I worked with lambdas more regularly, this might be obvious but at the moment it's not so even a link to aws docs would be helpful.

👍  I will add some instructions

> Which type of operation were you optimizing performance for with Vector here? Is it the loop in findConsentEmails that will be faster vs using a List?

List or Vector would have been fine in this scenario. The size and frequency the collections are being manipulated mean there's no need to worry about optimisation here. I often pick vector over list since all the operations have 'effectively constant' complexity according to the doc.

PR #52, opened 2020-04-03
Formstack SAR Lamdba

laurenemms:
This is part 1 of the Formstack SAR lambda.

This lambda is in charge of:
1. Initiating a subject access request
2. Checking the status of a running subject access requests (performed by lambda coming in next PR)

I have tested the status checking in code; however, I can't test the initiate method yet since the FormstackPerformSarHandler does not yet exist.

### DIFF
--- a/formstack-baton-requests/README.md
+++ b/formstack-baton-requests/README.md
@@ -1,0 +1,11 @@
+# Formstack Baton Requests
+
+Formstack Baton Requests is concerned with making Subject Access Requests to Formstack.
+
+## FormstackSarHandler
+The Formstack SAR Lambda is called from [Baton](https://github.com/guardian/baton) with two types of requests.
+1. An **initiate** request generates a UUID, which is then used by Baton to make subsequent status checks, and triggers the FormstackPerformSarLambda, which performs the requests to Formstack.
+2. A **status** request makes calls to S3 to check if an object exists in the results bucket with a completed or failed path. Otherwise, if an object exists in the failed path, a Failed response is returned to Baton. If an object exists in the completed path, a Completed response is returned to Baton with the results location. If no object exists for a given ID in neither the completed path nor the failed path, then a Pending response is returned to Baton.
+
+## FormstackPerformSarLambda 
+**COMING SOON**
--- a/formstack-baton-requests/build.sbt
+++ b/formstack-baton-requests/build.sbt
@@ -8,6 +8,9 @@ val circeVersion = "0.13.0"
 val amazonSdkVersion = "1.11.755"
 
 libraryDependencies ++= Seq(
+  "com.amazonaws" % "aws-lambda-java-core" % "1.2.0",
+  "com.amazonaws" % "aws-java-sdk-s3" % amazonSdkVersion,
+  "com.amazonaws" % "aws-java-sdk-lambda" % amazonSdkVersion,
   "com.typesafe.scala-logging" %% "scala-logging" % "3.9.0",
   "org.jlib" % "jlib-awslambda-logback" % "1.0.0",
   "io.circe" %% "circe-core" % circeVersion,
@@ -17,10 +20,6 @@ libraryDependencies ++= Seq(
 )
 
 scalacOptions += "-Ypartial-unification"
-
-addCompilerPlugin(
-  "org.scalamacros" % "paradise" % "2.1.1" cross CrossVersion.full
-)
 
 assemblyJarName := "main.jar"
 
--- a/formstack-baton-requests/build.sbt
+++ b/formstack-baton-requests/build.sbt
@@ -8,6 +8,12 @@ val circeVersion = "0.13.0"
 val amazonSdkVersion = "1.11.755"
 
 libraryDependencies ++= Seq(
+  "com.typesafe.scala-logging" %% "scala-logging" % "3.9.0",
+  "org.jlib" % "jlib-awslambda-logback" % "1.0.0",
+  "io.circe" %% "circe-core" % circeVersion,
+  "io.circe" %% "circe-parser" % circeVersion,
+  "io.circe" %% "circe-generic" % circeVersion,
+  "org.scalatest" %% "scalatest" % "3.0.5" % "test"
 )
 
 scalacOptions += "-Ypartial-unification"
--- a/formstack-baton-requests/build.sbt
+++ b/formstack-baton-requests/build.sbt
@@ -1,0 +1,32 @@
+
+name := "formstack-baton-requests"
+
+version := "0.1"
+
+scalaVersion := "2.12.8"
+val circeVersion = "0.13.0"
+val amazonSdkVersion = "1.11.755"
+
+libraryDependencies ++= Seq(
+)
+
+scalacOptions += "-Ypartial-unification"
+
+addCompilerPlugin(
+  "org.scalamacros" % "paradise" % "2.1.1" cross CrossVersion.full
+)
+
+assemblyJarName := "main.jar"
+
+assemblyMergeStrategy in assembly := {
+  case PathList("module-info.class") => MergeStrategy.discard
+  case x =>
+    val oldStrategy = (assemblyMergeStrategy in assembly).value
+    oldStrategy(x)
+}
+
+enablePlugins(RiffRaffArtifact)
+riffRaffPackageType := assembly.value
+riffRaffUploadArtifactBucket := Option("riffraff-artifact")
+riffRaffUploadManifestBucket := Option("riffraff-builds")
+riffRaffArtifactResources += (file("cloud-formation.yaml") -> "formstack-baton-requests-cfn/cloud-formation.yaml")
--- a/formstack-baton-requests/riff-raff.yaml
+++ b/formstack-baton-requests/riff-raff.yaml
@@ -1,0 +1,18 @@
+stacks: [identity]
+
+regions: [eu-west-1]
+
+deployments:
+  formstack-baton-requests-cfn:
+    type: cloud-formation
+    parameters:
+      templatePath: cloud-formation.yaml
+
+  formstack-baton-requests:
+    type: aws-lambda
+    parameters:
+      bucket: identity-lambda
+      functionNames: [FormstackSarLambda-]
+      fileName: main.jar
+      prefixStack: false
+    dependencies: [formstack-baton-requests-cfn]
--- a/formstack-baton-requests/riff-raff.yaml
+++ b/formstack-baton-requests/riff-raff.yaml
@@ -12,7 +12,7 @@ deployments:
     type: aws-lambda
     parameters:
       bucket: identity-lambda
-      functionNames: [FormstackSarLambda-]
+      functionNames: [formstack-baton-sar-lambda-]
       fileName: main.jar
       prefixStack: false
     dependencies: [formstack-baton-requests-cfn]
--- a/formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/BatonModels.scala
+++ b/formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/BatonModels.scala
@@ -1,0 +1,24 @@
+package com.gu.identity.formstackbatonrequests
+
+object BatonModels {
+
+  sealed trait BatonTaskStatus
+  case object Pending extends BatonTaskStatus
+  case object Completed extends BatonTaskStatus
+  case object Failed extends BatonTaskStatus
+
+  sealed trait SarRequest
+
+  case class SarInitiateRequest(subjectEmail: String, dataProvider: String) extends SarRequest
+  case class SarStatusRequest(initiationReference: String) extends SarRequest
+  case class SarPerformRequest(initiationReference: String, subjectEmail: String) extends SarRequest
+
+  sealed trait SarResponse
+
+  case class SarInitiateResponse(initiationReference: String) extends SarResponse
+  case class SarStatusResponse(
+    status: BatonTaskStatus,
+    resultLocations: Option[List[String]] = None,
+    message: Option[String] = None
+  ) extends SarResponse
+}
--- a/formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/CirceCodecs.scala
+++ b/formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/CirceCodecs.scala
@@ -1,0 +1,47 @@
+package com.gu.identity.formstackbatonrequests
+
+import com.gu.identity.formstackbatonrequests.BatonModels.{BatonTaskStatus, Completed, Failed, Pending, SarInitiateRequest, SarInitiateResponse, SarPerformRequest, SarRequest, SarResponse, SarStatusRequest, SarStatusResponse}
+import io.circe.{Decoder, Encoder, Json, JsonObject}
+import io.circe.generic.auto._
+import io.circe.syntax._
+
+object circeCodecs {
+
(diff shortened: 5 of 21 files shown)


formstack-baton-requests/README.md
```diff
@@ -0,0 +1,11 @@
+# Formstack Baton Requests
+
+Formstack Baton Requests is concerned with making Subject Access Requests to Formstack.
+
+## FormstackSarHandler
+The Formstack SAR Lambda is called from [Baton](https://github.com/guardian/baton) with two types of requests.
+1. An **initiate** request generates a UUID, which is then used by Baton to make subsequent status checks, and triggers the FormstackPerformSarLambda, which performs the requests to Formstack.
+2. A **status** request makes calls to S3 to check if an object exists in the results bucket with a completed or failed path. Otherwise, if an object exists in the failed path, a Failed response is returned to Baton. If an object exists in the completed path, a Completed response is returned to Baton with the results location. If no object exists for a given ID in neither the completed path nor the failed path, then a Pending response is returned to Baton.
```
tylr-mk, line 8: ✅

formstack-baton-requests/build.sbt
```diff
@@ -8,6 +8,12 @@ val circeVersion = "0.13.0"
 val amazonSdkVersion = "1.11.755"
 
 libraryDependencies ++= Seq(
```
tylr-mk, line 10: ✅ When I saw the empty call in commit f7012e3... I had my fingers crossed this was what you were doing. Excellent. Thanks

tylr-mk: Interesting that it's attached this comment here  and not the next commit where I made it 🤔

formstack-baton-requests/riff-raff.yaml
```diff
@@ -0,0 +1,18 @@
...
+      templatePath: cloud-formation.yaml
+
+  formstack-baton-requests:
+    type: aws-lambda
+    parameters:
+      bucket: identity-lambda
+      functionNames: [FormstackSarLambda-]
+      fileName: main.jar
+      prefixStack: false
+    dependencies: [formstack-baton-requests-cfn]
```
tylr-mk, line 18: ✅ my guess is all these general configs are all legit and knowledge based.  Going to be one of those times I'd like  to double check with @pvighi or someone else

formstack-baton-requests/build.sbt
tylr-mk: I think the comments aren't attached to commits but to lines and GitHub just provides them where that line is rendered and isn't out of date. Interesting choice.

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/BatonModels.scala
```diff
@@ -0,0 +1,24 @@
+package com.gu.identity.formstackbatonrequests
+
+object BatonModels {
+
+  sealed trait BatonTaskStatus
+  case object Pending extends BatonTaskStatus
+  case object Completed extends BatonTaskStatus
+  case object Failed extends BatonTaskStatus
```
tylr-mk, line 8: ✅ Which encapsulate all the statuses I expect from the README

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/CirceCodecs.scala
```diff
@@ -0,0 +1,47 @@
...
+
+import com.gu.identity.formstackbatonrequests.BatonModels.{BatonTaskStatus, Completed, Failed, Pending, SarInitiateRequest, SarInitiateResponse, SarPerformRequest, SarRequest, SarResponse, SarStatusRequest, SarStatusResponse}
+import io.circe.{Decoder, Encoder, Json, JsonObject}
+import io.circe.generic.auto._
+import io.circe.syntax._
+
+object circeCodecs {
+
+  implicit val sarRequestDecoder: Decoder[SarRequest] =
+    Decoder.instance[SarRequest] { cursor =>
```
tylr-mk, line 11: Scala funniness... I'll get there.

tylr-mk: This is the implicit function definition?

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/CirceCodecs.scala
```diff
@@ -0,0 +1,47 @@
...
+      cursor.downField("action").as[String].flatMap {
+        case "status" => cursor.as[SarStatusRequest]
+        case "initiate" => cursor.as[SarInitiateRequest]
```
tylr-mk, line 14: So this is the Baton pattern, to have two functions:

1. start
2. poll for status

?

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/FormstackConfig.scala
```diff
@@ -0,0 +1,25 @@
...
+
+case class SarLambdaConfig (
+  resultsBucket: String,
+  resultsPath: String,
+  performSarFunctionName: String
+)
+
+object FormstackConfig {
+
+  private def getEnvironmentVariable(env: String): Option[String] =
```
tylr-mk, line 11: in Python, the pattern is just to call the os env functions directly. Is there a reason for providing this as a private method?

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/FormstackConfig.scala
```diff
@@ -0,0 +1,25 @@
...
+    Option(System.getenv(env))
+
+  def getSarHandlerConfig: SarLambdaConfig = {
+    (for {
+      resultsBucket <- getEnvironmentVariable("resultsBucket")
+      resultsPath <- getEnvironmentVariable("resultsPath")
+      performSarFunctionName <- getEnvironmentVariable("performSarFunctionName")
```
tylr-mk: So when the lambda is invoked, the service has resultsBucket and resultsPath as ENV Variables?

If that is the case, again a pattern but upper snake casing is convention. But I'll leave that to you to decide. This is likely a pattern from other lambda Baton things.

tylr-mk: If they are on the ENV, how do they get there?

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/CirceCodecs.scala
```diff
@@ -44,4 +44,12 @@ object circeCodecs {
     Encoder.instance { psr =>
       addAdditionalFields(psr.asJsonObject, "SAR", "perform")
     }
+
+  /** encoder used for test run. */
+  implicit val sarRequestEncoder: Encoder[SarRequest] = Encoder.instance {
+    case ir: SarInitiateRequest =>
+      addAdditionalFields(ir.asJsonObject, "SAR", "initiate")
```
tylr-mk, line 51: This actual case logic being on the right hand side blows my tiny mind

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/FormstackSarHandler.scala
```diff
@@ -0,0 +1,47 @@
...
+
+    val performSarRequest = SarPerformRequest(
+      initiationReference,
+      request.subjectEmail
+    )
+
+    logger.info(s"invoking FormstackPerformSarLambda with initiation reference: $initiationReference")
+    lambdaClient.invokeLambda(performSarRequest, sarHandlerConfig)
+      .map(_ => SarInitiateResponse(initiationReference))
+  }
```
tylr-mk, line 25: I know this is Scala but unpacking a map and `_` magic feels like overkill. I'm not best placed to have an opinion

But whats different to

```suggestion
    lambdaClient.invokeLambda(performSarRequest, sarHandlerConfig)
    SarInitiateResponse(initiationReference)
  }
```

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/FormstackSarHandler.scala
```diff
@@ -0,0 +1,47 @@
...
+
+  private def status(initiationReference: String): Either[Throwable, SarStatusResponse] = {
+    logger.info(s"checking Formstack SAR status for initiation reference: $initiationReference")
+    s3Client.checkForResults(initiationReference, sarHandlerConfig).map {
+      case S3CompletedPathFound(resultLocations) =>
+        logger.info(s"SAR completed: completed SAR results for initiation reference $initiationReference found in s3: $resultLocations")
```
tylr-mk, line 31: resultsLocation a full S3 URI?

tylr-mk (commented):
I've got all the way to 3rd commit and the actual application logic in formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/aws.scala

... will resume when I can.

Lots of good work gone into making this reviewable for me. Meaningfully, that's meant easier review, and moving me from ~72% willing to approve without further guidance to ~86%.

tylr-mk:
Other meaningful things to say, this is great. Thanks.

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/CirceCodecs.scala
pvighi: This is defining an implicit Decoder from json to to SarRequest.

The way it creates the instance is by calling  Decoder.instance that takes function as a parameter:
https://github.com/circe/circe/blob/master/modules/core/shared/src/main/scala/io/circe/Decoder.scala#L500

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/FormstackConfig.scala
pvighi: they are set in the cloudformation when you create the lamba:
https://github.com/guardian/identity-processes/pull/52/files#diff-50b9fa1f83b8f0c93b951e86b27c3077R78

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/FormstackSarHandler.scala
pvighi: Well this function needs to return  `Either[Throwable, SarInitiateResponse]` so we cannot just return a `SarInitiateResponse`. 
InvokeLambda returns an `Either[Throwable, InvokeResult]` so mapping gets the correct return type.
This mapping is equivalent to just doing:
```scala
lambdaClient.invokeLambda(performSarRequest, sarHandlerConfig) match {
case Right(_) => Right(SarInitiateResponse(initiationReference))
case Left(error) => Left(error)
```
the `_` basically is there because it looks like in this case we don't really care what the successful response was, we are satisfied with it not returning an error

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/FormstackSarHandler.scala
```diff
@@ -0,0 +1,47 @@
...
+  private def initiate(request: SarInitiateRequest): Either[Throwable, SarInitiateResponse] = {
+    val initiationReference = randomUUID.toString
+
+    val performSarRequest = SarPerformRequest(
+      initiationReference,
+      request.subjectEmail
+    )
+
+    logger.info(s"invoking FormstackPerformSarLambda with initiation reference: $initiationReference")
+    lambdaClient.invokeLambda(performSarRequest, sarHandlerConfig)
```
pvighi, line 23: why does this lambda call another lambda?

laurenemms: This is how all the lambda's for baton have been working (excluding the Baton lambda which can use a query ID returned by Athena). 1st lambda only deals with status checking (in S3) and initiating the lambda that actually performs with SAR. Then the second one performs the actual SAR and writes to S3.

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/FormstackSarHandler.scala
laurenemms: 👍 explained better than I could have

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/CirceCodecs.scala
tylr-mk: I meant specifically the syntax that allows `Decoder.... NOTHING { ??? cursor ??? =>`

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/FormstackConfig.scala
tylr-mk: So is it appropriate to ask that they are named like system ENV VARs

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/FormstackConfig.scala
laurenemms: I moved this into a function mostly because it's used a number of times (and when I add the second lambda to this project, it'll be called even more). It's not saving me writing that much code but I thought maybe it made it more explicit what wrapping this in an Option and System.getenv are doing

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/CirceCodecs.scala
laurenemms: Yes! Actually 3, the third will be 'perform' and is actioned by the next lambda I'm working on. The "perform" action is only triggered by a call from the first lambda to the second lambda, and is not called from Baton directly

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/FormstackConfig.scala
laurenemms: I've renamed them to upper snake case :)

formstack-baton-requests/build.sbt
```diff
@@ -0,0 +1,41 @@
...
+  "io.circe" %% "circe-core" % circeVersion,
+  "io.circe" %% "circe-parser" % circeVersion,
+  "io.circe" %% "circe-generic" % circeVersion,
+  "org.scalatest" %% "scalatest" % "3.0.5" % "test"
+)
+
+scalacOptions += "-Ypartial-unification"
+
+addCompilerPlugin(
+  "org.scalamacros" % "paradise" % "2.1.1" cross CrossVersion.full
```
pvighi: minor: what is this for ?

pvighi (approved):
👍

formstack-baton-requests/build.sbt
laurenemms: I've removed this now. It gave me access to @JsonCodec in circe but I don't need it anymore

PR #53, opened 2020-04-09
Add dynamo status tables to Formstack-Baton-Requests

laurenemms:
The aim of this PR is to create two tables (in PROD only) as part of the work for the Formstack SAR lambda. Formstack returns a paginated response for form submissions, which can take hours to iterate through. For this reason, we're going to get all these submissions once only, write them to dynamo and then only query formstack to update this table.

1. formstack-submission-ids 
This has an 'email' partition key (so we can search for a given email address) and 'submission id' as a sort key.

2. formstack-submissions-last-updated
This table is to store stateful metadata for the above-mentioned table. The reason for this table is to store when formstack-submission-ids was last updated. With this, we can make calls to Formstack's API to retrieve submissions SINCE the lastUpdated date and just update the formstack-submission-ids table with those submissions.

### DIFF
--- a/formstack-baton-requests/cloud-formation.yaml
+++ b/formstack-baton-requests/cloud-formation.yaml
@@ -19,6 +19,9 @@ Parameters:
     Description: Subnets to use in VPC
     Type: CommaDelimitedList
 
+Conditions:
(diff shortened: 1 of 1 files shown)


ravimbhatt (approved):
LGTM

PR #55, opened 2020-04-23
Implement perform sar lambda

laurenemms:
This is the last PR for the Formstack SAR work.

There's a bit of refactoring in it but the main goal of this PR is to create the functionality that performs the actual SAR and ties the the process of updating and retrieving from Dynamo, collecting additional data from Formstack, and writing to S3. 

I've also created some tests and a stub lambda for running in CODE.

Tested locally and will test in CODE tomorrow while this is in review.


formstack-baton-requests/project/plugins.sbt
```diff
@@ -1,3 +1,4 @@
 addSbtPlugin("com.gu" % "sbt-riffraff-artifact" % "1.0.0")
 
-addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.10.0-RC1")
\ No newline at end of file
+addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.10.0-RC1")
```
ravimbhatt: do we need this?

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/FormstackSarService.scala
```diff
@@ -77,4 +86,58 @@ object FormstackSarService extends FormstackSar with LazyLogging {
     else
       decode[FormstackFormSubmissionsResponse](response.body)
   }
+
+  private def submissionsById(
```
ravimbhatt: this method is called submissionsById but returns FormstackSubmissionResponse. Naming is hard, i think we should try and give it a better name.

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/FormstackSarService.scala
```diff
@@ -77,4 +86,58 @@ object FormstackSarService extends FormstackSar with LazyLogging {
...
+
+      if(!response.is2xx) {
+        logger.error(response.body)
+      }
+
+      decode[FormstackSubmissionResponse](response.body)
+    }
+  }
+
+  private def retrieveSubmissionLabels(
```
ravimbhatt: same as above. may be getSubmissionQuestionAnswers?

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/FormstackSarService.scala
```diff
@@ -6,10 +6,12 @@ import io.circe.parser.decode
...
 import io.circe.generic.JsonCodec
+import cats.implicits._
 
 trait FormstackSar {
   def accountFormsForGivenPage(page: Int, accountToken: FormstackAccountToken): Either[Throwable, FormstackFormsResponse]
   def formSubmissionsForGivenPage(page: Int, formId: String, minTime: SubmissionTableUpdateDate, encryptionPassword: String, accountToken: FormstackAccountToken): Either[Throwable, FormstackFormSubmissionsResponse]
+  def submissionData(submissionIdEmails: List[SubmissionIdEmail], config: PerformSarLambdaConfig): Either[Throwable, List[FormstackSubmissionQuestionAnswer]]
 }
 /* Codecs for decoding accountFormsForGivenPage response */
 @JsonCodec case class FormstackForm(id: String)
```
ravimbhatt: would it make it easy if we removed Formstack from all these case classes? They belong to Formstak which is claer from the class and package name,

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/aws/S3Client.scala
```diff
@@ -13,9 +14,13 @@ case class S3CompletedPathFound(resultLocations: List[String]) extends S3StatusR
 case class S3FailedPathFound() extends S3StatusResponse
```
ravimbhatt: again this class is called S3Client. I personally think that the methods should not have s3 in them. we could read it as s3Client.S3FailedPathFound or s3Client.FailedPathFound

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/aws/S3Client.scala
```diff
@@ -64,4 +69,43 @@ object S3 extends S3Client with LazyLogging {
...
+      (meta ::: responses).mkString("\n")
+    }.mkString("\n-------------------------\n")
+  }
+
+  override def writeSuccessResult(
+    initiationId: String,
+    results: List[FormstackSubmissionQuestionAnswer],
+    config: PerformSarLambdaConfig): Either[Throwable, S3WriteSuccess] = {
+    val resultsPath = s"${config.resultsPath}/$initiationId/completed"
+    if (results.nonEmpty) {
```
ravimbhatt: i think we could replace this block with: 


```
 val filePath, contents =  if (results.nonEmpty) {
       logger.info("Writing SAR result to s3.")
       val filePath = s"$resultsPath/formstackSarResponse"
      val contents = formatResults(results)
      filePath, contents 
    } else {
      logger.info(s"No results found for request $initiationId. Creating NoResultsFoundForUser object.")
      val filePath = s"$resultsPath/noResultsFoundForUser"
      filePath, ""
    }

      writeToS3(config.resultsBucket, filePath, contents)
```

or better still if formatResults return empty string on the list being empty. so if condition check now return just the filepath and we could use writeToS3 as 
` 
writeToS3(config.resultsBucket, filePath, formatResults(contents))
 `

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/FormstackSarService.scala
laurenemms: agree - done

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/FormstackSarService.scala
laurenemms: agree - done

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/aws/S3Client.scala
laurenemms: agree - done

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/FormstackSarService.scala
laurenemms: done!

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/aws/S3Client.scala
laurenemms: Ah nice - thank you.

ravimbhatt (approved):
LGTM

PR #57, opened 2020-04-28
Fix pages bug, modify submissions decoder and add alarms 

laurenemms:
This PR does a few things that I realised needed fixing when running the script:

- Move codecs into CirceCodecs file and add a custom decoder for FormSubmissions. Due to some inconsistencies in the Formstack API responses, sometimes data can be decoded to a Map[String, ResponseValue] but sometimes data is an empty array, which cannot be decoded to a Map[String, ResponseValue]. These responses are still valid, but they don't include data, so we can return Map.empty. Also added tests for this.

- Modified email regex to allow the end of email addresses to be between 2-6 letters long as I saw a few that were 5 letters long. Hopefully this will be good enough.

- Change recursive logic from (count + FormstackSarService.resultsPerPage) <= response.total to count < response.total as it was missing the last page.

- Added an alarm for errors.

### DIFF
--- a/formstack-baton-requests/cloud-formation.yaml
+++ b/formstack-baton-requests/cloud-formation.yaml
@@ -41,6 +41,14 @@ Conditions:
   IsProd: !Equals [ !Ref Stage, PROD ]
 
 Resources:
+  TopicPagerDutyAlerts:
+    Type: AWS::SNS::Topic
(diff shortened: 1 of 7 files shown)


ravimbhatt (approved):
LGTM

PR #58, opened 2020-04-29
update eventbrite api call

QuarpT:
Eventbrite are changing their API. We now have to pass an organisation ID to a slightly different path. The response body and parameters are the same.

The API call is used to get users who have ticked a 'hear more from guardian events' button. We hit the API twice with 
- an events organisation id and token
- a masterclasses organisation id and token


dominickendrick (approved):
Looks good !

PR #59, opened 2020-04-30
Formstack RER Lambda

laurenemms:
### What is this?
This PR deals with Right to Erasure Requests to Formstack.

As with the Formstack SAR process, this includes 2 lambdas, one for status checking and initiating the 'Perform' lambda, and one for performing the right to erasure process and writing to certain status paths in S3 (for the first lambda to check). 

Other than adding those 2 lambdas, the models that come with them and the circe codecs. This PR also does the following:
- Refactors some of the config, classes and code to make it usable by both SAR and RER processes. This includes the creation of the `DynamoUpdateService` which is now called at the beginning of both the SAR and RER processes to ensure the Dynamo table is always up to date.
- Additional method for deleting submissions from Formstack
- Additional method for deleting submissions from Dynamo
- Add a check to make sure that whatever type of request is called "SAR" or "RER" includes the correct request type in the request. Let me know if anyone thinks we need more safety checks around this.

### Testing
Tested locally by:
- Submitted a response to a form in Formstack with my own email address
- Updated the dynamo table to include new requests
- This then successfully deleted when I ran the perform RER lambda (both from Formstack and Dynamo)

Also tested in CODE.


formstack-baton-requests/cloud-formation.yaml
```diff
@@ -8,8 +8,8 @@ Parameters:
       - PROD
       - CODE
     Default: CODE
-  SarResultsBucket:
-    Description: Bucket where sar results are uploaded to
+  ResultsBucket:
```
ravimbhatt, line 11: do we store files for both SARs and RERs here? we identify them by just filesnames? No problems accidentally deleting all files in this bucket?

formstack-baton-requests/cloud-formation.yaml
```diff
@@ -274,6 +274,168 @@ Resources:
...
+
+  PerformFormstackRerLambda:
+    Type: AWS::Lambda::Function
+    Properties:
+      Description: Performs RER to Formstack, writing status updates to S3
+      FunctionName:
+        !Sub formstack-baton-perform-rer-lambda-${Stage}
+      Code:
+        S3Bucket: identity-lambda
+        S3Key: !Sub identity/${Stage}/formstack-baton-requests/main.jar
```
ravimbhatt, line 465: main.jar? usually we have projectname.jar. not a big deal though.

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/DynamoUpdateService.scala
```diff
@@ -0,0 +1,81 @@
...
+import com.gu.identity.formstackbatonrequests.circeCodecs.{Form, FormSubmission}
+import com.typesafe.scalalogging.LazyLogging
+
+case class DynamoUpdateService(
+  formstackClient: FormstackRequestService,
+  dynamoClient: DynamoClient,
+  config: PerformLambdaConfig) extends LazyLogging {
+
+  def submissionsWithEmailAndAccount(submissions: List[FormSubmission], accountNumber: Int): List[SubmissionIdEmail] = {
+    val emailReg = """(?i)\b[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,4}\b""".r
```
ravimbhatt: didn't we change this to 6 chars for the chars after the dot?

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/DynamoUpdateService.scala
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  private def writeSubmissions(form: Form, lastUpdate: SubmissionTableUpdateDate, submissionPage: Int = 1, token: FormstackAccountToken): Either[Throwable, Unit] = {
+    formstackClient.formSubmissionsForGivenPage(submissionPage, form.id, lastUpdate, config.encryptionPassword, token) match {
+      case Left(err) => skipDecryptionError(err)
+      case Right(response) =>
+        logger.info(s"Received page $submissionPage of submissions out of ${response.pages} pages for form ${form.id}.")
+        val submissionsIdsWithEmails = submissionsWithEmailAndAccount(response.submissions, token.account)
+        logger.info(s"Writing ${submissionsIdsWithEmails.length} submission id(s) and emails to Dynamo")
+        dynamoClient.writeSubmissions(submissionsIdsWithEmails, config.bcryptSalt, config.submissionTableName) match {
```
ravimbhatt, line 39: may be creating 2-3 small functions here would make this more readable. too many right left in this function.

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/DynamoUpdateService.scala
```diff
@@ -0,0 +1,81 @@
...
+    formstackClient.accountFormsForGivenPage(formsPage, token) match {
+      case Left(err) => Left(err)
+      case Right(response) =>
+        val forms = response.forms
+        val formResults = forms.map { form =>
+          logger.info(s"Processing results for form ${form.id}")
+          writeSubmissions(form, lastUpdate, token = token)
+        }
+        val errors = formResults.collect { case Left(err) => err }
+        if (errors.nonEmpty) {
```
ravimbhatt, line 63: could errors be empty and the response from a collect still be a Left?

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/DynamoUpdateService.scala
laurenemms: Oh no, maybe I lost the change.. Thanks for catching this. I'll make sure nothing else was lost from that PR

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/DynamoUpdateService.scala
laurenemms: I don't think so!

formstack-baton-requests/cloud-formation.yaml
laurenemms: I THINK it's okay as each one should have its own initiation reference (which will be in the path) but I've just made some changes just in case to add the requestType to the path in S3. It will definitely make it clearer 👍

formstack-baton-requests/cloud-formation.yaml
laurenemms: I just did this to keep it in line with the other lambdas that identity has in this project

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/DynamoUpdateService.scala
laurenemms: Yeah.. I'm struggled to find a good was of breaking this down, since it's recursive and it would mean passing so many parameters. Does anything obvious jump out to you?

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/aws/S3Client.scala
```diff
@@ -86,14 +97,19 @@ object S3 extends S3Client with LazyLogging {
...
-      logger.info("Writing SAR result to s3.")
-      s"$resultsPath/formstackSarResponse"
-    } else {
-      logger.info(s"No results found for request $initiationId. Creating NoResultsFoundForUser object.")
-      s"$resultsPath/noResultsFoundForUser"
+    requestType: BatonRequestType,
+    config: PerformLambdaConfig): Either[Throwable, S3WriteSuccess] = {
+
+    val filePath = requestType match {
+      case SAR if results.nonEmpty =>
```
ravimbhatt: we can have only one case here. We check for case SAR and use the results empty check to populate the last Option that goes into the generateResultsPath method.

ravimbhatt: or better still move this logic to generateResultsPath method.

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/aws/S3Client.scala
```diff
@@ -102,9 +118,17 @@ object S3 extends S3Client with LazyLogging {
   override def writeFailedResults(
     initiationId: String,
     err: String,
-    config: PerformSarLambdaConfig): Either[Throwable, S3WriteSuccess] = {
-    val filePath = s"${config.resultsPath}/$initiationId/failed/formstackSarFailed"
-    logger.info("Writing to failed path in s3.")
+    requestType: BatonRequestType,
+    config: PerformLambdaConfig): Either[Throwable, S3WriteSuccess] = {
+    val filePath = requestType match {
```
ravimbhatt: same as above, if we move creation of this Option object to generateResultsPath, we have one line and no match case.

ravimbhatt:
i think we are good, other than 1-2 small comments.

ravimbhatt (approved):
please merge if you have tested this in code.

PR #61, opened 2020-05-04
Refactor using packages

laurenemms:
Just breaking down some of the files into packages to make the code more approachable now that there is logic for both RERs and SARs in this project.


ravimbhatt (approved):
nice cleanup

PR #62, opened 2020-05-11
Fix supported providers for Formstack requests

laurenemms:
When I updated Baton to have a separate RER provider Formstack, I forgot to update the lambdas to support the new data provider.

This just extends the support to the FormstackRer data provider.

Tested in CODE.


ravimbhatt (approved):
+1

PR #63, opened 2020-05-12
Increase write capacity for Dynamo update

laurenemms:
When running the Dynamo table update locally the other day, I noticed that even though it's only updating since the last time the lambda ran, this sometimes comes extremely close to the lambda time limit (15 minutes). 

One way to improve the speed of the update is to increase the write capacity per second to  the submissions table Dynamo. This can be achieved by adjusting the provisioned throughput capacity depending on the resources being used. The problem with this approach is that it won't scale up until the target value is exceeded, and then it takes around 5 minutes for the actual increase to be in effect. If the target value isn't exceeded until the 11th minute, the lambda will never see the effect of this policy.

For this reason, I've included a step to increase the capacity manually at the start of every run, and reduce the capacity at the end. This is still not perfect, as it still takes 5 minutes for the change to come into effect. However, if there are a lot of submissions to be added to the table, at least the increased capacity will work after 5 minutes.

Hopefully this change will be enough... Otherwise, we might have to look into either:
- Separating the update from the rest of the SAR/RER process into it's own lambda and schedule the update daily so that there isn't a big gap between the times the table is updated
- Turning the lambdas in step functions with check conditions.

### DIFF
--- a/formstack-baton-requests/cloud-formation.yaml
+++ b/formstack-baton-requests/cloud-formation.yaml
@@ -249,6 +249,7 @@ Resources:
                 Action:
                   - dynamodb:BatchWriteItem
                   - dynamodb:Query
+                  - dynamodb:UpdateTable
                 Resource: !Sub arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/${SubmissionsTableName}
         - PolicyName: LastUpdatedTablePolicy
           PolicyDocument:
@@ -344,6 +345,7 @@ Resources:
                 Action:
                   - dynamodb:BatchWriteItem
                   - dynamodb:Query
+                  - dynamodb:UpdateTable
                 Resource: !Sub arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/${SubmissionsTableName}
         - PolicyName: LastUpdatedTablePolicy
           PolicyDocument:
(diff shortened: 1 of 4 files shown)


ravimbhatt:
> For this reason, I've included a step to increase the capacity manually at the start of every run, and reduce the capacity at the end. This is still not perfect, as it still takes 5 minutes for the change to come into effect. However, if there are a lot of submissions to be added to the table, at least the increased capacity will work after 5 minutes.

so this happens in baton before the lambda is called?

laurenemms:
> > For this reason, I've included a step to increase the capacity manually at the start of every run, and reduce the capacity at the end. This is still not perfect, as it still takes 5 minutes for the change to come into effect. However, if there are a lot of submissions to be added to the table, at least the increased capacity will work after 5 minutes.
> 
> so this happens in baton before the lambda is called?

No, this happens right at the start of the lambda. Still not ideal as will take 5 mins to come into effect. This morning I'm looking into actually turning this into a step function. I've got a feeling we'll have problems with how long this takes otherwise. But this might be an okay fix for now.

ravimbhatt:
> > > For this reason, I've included a step to increase the capacity manually at the start of every run, and reduce the capacity at the end. This is still not perfect, as it still takes 5 minutes for the change to come into effect. However, if there are a lot of submissions to be added to the table, at least the increased capacity will work after 5 minutes.
> > 
> > 
> > so this happens in baton before the lambda is called?
> 
> No, this happens right at the start of the lambda. Still not ideal as will take 5 mins to come into effect. This morning I'm looking into actually turning this into a step function. I've got a feeling we'll have problems with how long this takes otherwise. But this might be an okay fix for now.

okay. if it takes 5 mins to change the limit then we have spent 10 mins out of 15 just doing that. 

It changes straight away, it just takes 5 minutes to kick in. So for the first 5 minutes, it will continue updating the table but at a slower pace, then after 5 minutes it will do it faster

ravimbhatt (approved):

PR #65, opened 2020-05-21
Add new Baton account invoke role

laurenemms:
### Why are you making this change?
This adds permissions for the Formstack SAR and RER init lambdas to be invoked from the new Baton AWS account.

The identity invoke role will be deleted once the migration is completed.

### What background information is there? (e.g. an [ADR](https://github.com/guardian/ophan-data-lake/blob/master/decisions/template.md), Trello card, PR, or Issue)

https://trello.com/c/oFgWKRhL/858-add-role-to-all-lambdas-for-new-baton-aws-account

### DIFF
--- a/formstack-baton-requests/cloud-formation.yaml
+++ b/formstack-baton-requests/cloud-formation.yaml
@@ -36,6 +36,8 @@ Parameters:
   SubmissionsTableName:
     Type: String
     Default: formstack-submission-ids
+  BatonAccountId:
+    Type: String
 
 Conditions:
   IsProd: !Equals [ !Ref Stage, PROD ]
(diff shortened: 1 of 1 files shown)


tylr-mk (approved):
The word *root* causes me concern, apprehension.

That being said, this looks good. So ✅ 

Before merging please answer the questions: what root means and does in this  case. And also link  to another case in this PR comment.

laurenemms:
> The word _root_ causes me concern, apprehension.
> 
> That being said, this looks good. So ✅
> 
> Before merging please answer the questions: what root means and does in this case. And also link to another case in this PR comment.

Root basically just means giving the permissions to the Baton AWS account itself. This allows cross-account invocation. Root user ARNs are often used in the Principal elements of policies.

The Principal element's specify who can assume the role (more info [here](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_principal.html)).

Here are a couple of other example from ophan-data-lake where we are using root user arns to permission cross-account functionality:
- https://github.com/guardian/ophan-data-lake/blob/6fabb528e8c701233d545b378408008c20bc5094/cloudformation/cross-account-athena-access.yaml#L30
- https://github.com/guardian/ophan-data-lake/blob/877497e35bf2ebae84e707f05d052db32aea4ed8/baton-lambda/cfn.yaml#L79

Hope this explains it, let me know if you need any more details. I'll post this in the other PR as well.

PR #67, opened 2020-07-22
Fix lambda to not break when continuation and has_more_items fields are absent in the Eventbrite response

rebecca-thompson:
## What does this change?
The Eventbrite API does not always send a response with the pagination.continuation and pagination.has_more_items fields. This PR adapts our current model to allow for these fields to be absent, provided the page count is less than the current page number (ie, there are no more pages to process).

## How to test
There are instructions in the README.md on how to run this locally in DEV

We should no longer have failing Eventbrite consent jobs when pagination.continuation and pagination.has_more_items are absent for valid reasons.

## Have we considered potential risks?
Yes

### DIFF
--- a/eventbrite-consents/README.md
+++ b/eventbrite-consents/README.md
@@ -29,7 +29,9 @@ Run with SBT
 ```
 export idapiHost=not_used_in_debug
 export idapiAccessToken=not_used_in_debug
+export masterclassesOrganisation=set_this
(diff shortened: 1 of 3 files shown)


rebecca-thompson:
[Fix the EventbriteConsentsLambda](https://trello.com/c/lLTQuOvx/2278-fix-the-eventbriteconsentslambda)

PR #70, opened 2020-08-26
Update alarm names and send emails via AWS SNS topic rather than Pagerduty

lmath:
## What does this change?

We want to make it so that when alerts hit our inboxes, they have some useful information about why the alert has been triggered and what codebase it belongs to.

When we configure alerts, there is usually

-  a description associated with the alert.
- we can also define a readable name for the alert.
- we know what repo the alert belongs to

This information doesn't come through in the PagerDuty alerts and though I'm sure we could make the PagerDuty alerts more informative we seem to just be using PagerDuty to send emails. So this PR
- sends emails via the existing SNS topic rather than pagerduty
- adds alarm names
- updates alarm descriptions

The email address that should receive the alerts should be specified as a cloudformation parameter so to release this
#### in CODE
- [x] try adding in an email topic and test alert in CODE, and see that the expected name and description arrive as an email (& tidy up after)
- [x] manually update the cloudformation & specify alarm email address param - payment failure lambda - then check it is possible to deploy the build with the changes
- [x] manually update the cloudformation & specify alarm email address param - eventbrite consents lambda - then check it is possible to deploy the build with the changes
- [x] manually update the cloudformation & specify alarm email address param - formstack baton consents lambda - then check it is possible to deploy the build with the changes
- [x] manually update the cloudformation & specify alarm email address param - formstack consents lambda - then check it is possible to deploy the build with the changes

#### in PROD (after PR approval)
- [x] manually update the cloudformation & specify alarm email address param - payment failure lambda
- [x] manually update the cloudformation & specify alarm email address param - eventbrite consents lambda
- [x] manually update the cloudformation & specify alarm email address param - formstack baton consents lambda
- [x] manually update the cloudformation & specify alarm email address param - formstack consents lambda
- [x] check that each these lambdas have deployed successfully after the merge

### DIFF
--- a/payment-failure/cloud-formation.yaml
+++ b/payment-failure/cloud-formation.yaml
@@ -46,13 +46,6 @@ Resources:
       Subscription:
         - Endpoint: !Ref 'AlarmEmailAddress'
           Protocol: email
-  TopicPagerDutyAlerts:
-    Type: AWS::SNS::Topic
-    Properties:
-      DisplayName: PagerDutyTopic
-      Subscription:
-      - Endpoint: https://events.pagerduty.com/adapter/cloudwatch_sns/v1/96fdc0179acb4c5db3e059d775ea6a9e
-        Protocol: https
   PaymentFailureLambdaRole:
     Type: AWS::IAM::Role
     Properties:
--- a/payment-failure/cloud-formation.yaml
+++ b/payment-failure/cloud-formation.yaml
@@ -152,8 +152,7 @@ Resources:
     Type: AWS::CloudWatch::Alarm
     Properties:
       AlarmName: !Sub 'payment-failure-lambda-${Stage} Message added to payment failure dead letter queue'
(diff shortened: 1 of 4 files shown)


payment-failure/cloud-formation.yaml
```diff
@@ -148,6 +151,7 @@ Resources:
   PaymentFailureDeadLetterQueueAlert:
     Type: AWS::CloudWatch::Alarm
     Properties:
+      AlarmName: !Sub 'payment-failure-lambda-${Stage} Message added to payment failure dead letter queue'
       AlarmDescription: Alarm if payment failure dead letter queue grows beyond 1
         message
```
rebecca-thompson: is newline required?

rebecca-thompson (approved):

payment-failure/cloud-formation.yaml
lmath: I will remove it!

PR #74, opened 2020-10-29
Update Eventbrite consents lambda to handle invalid email addresses

rebecca-thompson:
## What does this change?
When a user submits an invalid email address, log the error rather than throwing a Runtime exception which terminates the application. 

We were getting alerts about lambda errors because a user had entered an invalid email address. We do not want this error to terminate the application as it means other emails in the batch do not get processed, so we log the error instead and throw an exception for any other scenario.

## How to test
There are unit tests.

Email consent batches should complete successfully if one email address in the batch is invalid and we should still be alerted to any other errors.

### DIFF
--- a/eventbrite-consents/src/main/scala/com/gu/identity/eventbriteconsents/clients/IdentityClient.scala
+++ b/eventbrite-consents/src/main/scala/com/gu/identity/eventbriteconsents/clients/IdentityClient.scala
@@ -33,13 +33,13 @@ class IdentityClient(idapiUrl: String, idapiAccessToken: String) extends LazyLog
         } yield if (error.description.contains("Bad email format")) {
           logger.error("Invalid email address, could not process consents: " + emailAddress)
         } else {
-          throwRuntimeException
+          throwRuntimeException()
         }
       }
-      case _ => throwRuntimeException
+      case _ => throwRuntimeException()
     }
 
-    def throwRuntimeException: Nothing = {
+    def throwRuntimeException(): Nothing = {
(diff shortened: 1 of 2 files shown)


eventbrite-consents/src/main/scala/com/gu/identity/eventbriteconsents/clients/IdentityClient.scala
```diff
@@ -16,6 +20,27 @@ class IdentityClient(idapiUrl: String, idapiAccessToken: String) {
...
+        } yield if (error.description.contains("Bad email format")) {
+          logger.error("Invalid email address, could not process consents: " + emailAddress)
+        } else {
+          throwRuntimeException
+        }
+      }
+      case _ => throwRuntimeException
+    }
+
+    def throwRuntimeException: Nothing = {
```
QuarpT: for functions with side effects it's idiomatic to use `()` in the signature, e.g. `throwRuntimeException()`

QuarpT (approved):
looks great

eventbrite-consents/src/main/scala/com/gu/identity/eventbriteconsents/clients/IdentityClient.scala
rebecca-thompson: Ah, good point. I'll change that

rebecca-thompson:
[Refactor Eventbrite consent lambda's error handling](https://trello.com/c/iWf3oL7F/2599-refactor-eventbrite-consent-lambdas-error-handling)

PR #75, opened 2020-11-17
Delay failure emails by 5 minutes to ensure emails are sent by braze

QuarpT:
## What does this change?

Delay the failure lambda ingest from SQS by 5 minutes to ensure emails sent via Braze are sent to users.

## Why?

Braze won't always send emails for newly created accounts.  (see https://www.braze.com/docs/help/best_practices/race_conditions/ )
Braze will return success responses even if the email fails to be sent.


johnduffell (approved):
great! although doesn't have to be the full 5 minutes, one would have been enough

QuarpT:
> great! although doesn't have to be the full 5 minutes, one would have been enough

👌  I'll put it down to two minutes to be safe

PR #77, opened 2021-09-13
Increase connection timeouts for Eventbrite API request

rebecca-thompson:
## What does this change?
This doubles the (default) socket connection and read timeouts for the Eventbrite consents Lambda. We are getting socket timeout exceptions in the Lambda logs because the request to the Eventbrite API is taking longer than the default timeout.

### DIFF
--- a/eventbrite-consents/src/main/scala/com/gu/identity/eventbriteconsents/clients/EventbriteClient.scala
+++ b/eventbrite-consents/src/main/scala/com/gu/identity/eventbriteconsents/clients/EventbriteClient.scala
@@ -27,8 +27,8 @@ class EventbriteClient {
         "changed_since" -> formattedLastRun,
         "continuation" -> continuationToken
       )
+      .timeout(2000, 10000)
       .asString
-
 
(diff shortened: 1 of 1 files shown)


rebecca-thompson:
[Triage tasks](https://trello.com/c/Ned8sVN8/3385-triage-tasks)

pvighi (approved):
looks good to me.
Given the 10 second timeout your are setting I assume the timeout on the lambda itself is not a problem since it seems to be set to [2 minutes](https://github.com/guardian/identity-processes/blob/main/eventbrite-consents/cloud-formation.yaml#L62)

rebecca-thompson:
> looks good to me.
> Given the 10 second timeout your are setting I assume the timeout on the lambda itself is not a problem since it seems to be set to [2 minutes](https://github.com/guardian/identity-processes/blob/main/eventbrite-consents/cloud-formation.yaml#L62)

Correct. The Java network errors are being thrown by the `EventbriteClient.findConsents` method in the logs, which is what calls the Eventbrite API

PR #85, opened 2021-12-16
upgrade sbt

QuarpT:
## What does this change?
Upgrade sbt to latest 1.5.7


pvighi (approved):
looks good to me if it all still works. I guess that there are fewer chances for dependency problems in smaller builds like this so you didn't need to disable coursier

PR #86, opened 2021-12-16
Remove dynamodb local

QuarpT:
## What does this change?

Replace sbt-dynamodb with dockerised dynamodb

## Why?

- sbt-dynamodb is no longer maintained
- @vlbee says sbt-dynamodb uses a vulnerable version of log4j

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 .idea/
+target/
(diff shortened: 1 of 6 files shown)


rebecca-thompson (commented):
Does the README.md need updating with instructions on how to run this locally?

pvighi (approved):
looks good to me

QuarpT:
> Does the README.md need updating with instructions on how to run this locally?

it looks like local dynamodb only gets used in the tests. For running locally I assume this project uses real DynamoDB. There aren't instructions for setting environment variables, so I don't know what tables it's supposed to use locally

PR #91, opened 2022-02-14
fix: try increasing page size so we make fewer total requests

ob6160:
## What does this change?
Increases the number of results we ask for per page to 100, hopefully reducing the possibility of a timeout as we are making fewer requests to Formstack

### DIFF
--- a/formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/FormstackService.scala
+++ b/formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/FormstackService.scala
@@ -20,7 +20,7 @@ case class FormstackDecryptionError(message: String) extends Throwable
 
 object FormstackService extends FormstackRequestService with LazyLogging {
 
-  val resultsPerPage = 25
+  val resultsPerPage = 100
 
   override def accountFormsForGivenPage(page: Int, accountToken: FormstackAccountToken): Either[Throwable, FormsResponse] = {
     val response = Http(s"https://www.formstack.com/api/v2/form.json")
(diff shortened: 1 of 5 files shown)


kenoir:
The results per page applies to the number of forms returned when paginating through an account AND the number of submissions per page when paginating through a form, I think if we decouple those values and request more submissions per page we might be able to address the issue of timeouts when paginating through huge forms.

Increasing the number of forms per page is likely to take each run of the `DynamoUpdateService` take longer as it processes more forms before going back and calling `updateSubmissionsTable` where it can check if it's running out of time. 

However increasing the number of submissions returned per form (as you suggest) will reduce the number of requests being made and quite likely speed things up.

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/FormstackService.scala
```diff
@@ -20,7 +20,7 @@ case class FormstackDecryptionError(message: String) extends Throwable
 
 object FormstackService extends FormstackRequestService with LazyLogging {
 
-  val resultsPerPage = 25
+  val resultsPerPage = 100
```
kenoir, line 23:
```suggestion
 val formResultsPerPage = 25
 val formSubmissionsPerPage = 100
```

kenoir:
```suggestion
  // Results returned by accountFormsForGivenPage
  val formResultsPerPage = 25
  // Results returned by formSubmissionsForGivenPage
  val submissionResultsPerPage = 100
```

ob6160:
> The results per page applies to the number of forms returned when paginating through an account AND the number of submissions per page when paginating through a form, I think if we decouple those values and request more submissions per page we might be able to address the issue of timeouts when paginating through huge forms.
> 
> Increasing the number of forms per page is likely to take each run of the `DynamoUpdateService` take longer as it processes more forms before going back and calling `updateSubmissionsTable` where it can check if it's running out of time.
> 
> However increasing the number of submissions returned per form (as you suggest) will reduce the number of requests being made and quite likely speed things up.

Thanks for the comment @kenoir , that makes a lot of sense. I'll change this to separate the two so that we can specifically load fewer pages of submissions and leave the number of forms as-is

ob6160 (commented):
thanks very much for finishing this off @kenoir, looks good - I think we just need to check this all works in CODE .. unless you've checked that already?

PR #92, opened 2022-04-04
Formstack Baton requests error handling 

ob6160:
## What does this change?
This PR makes the process make sure the submissions referenced in dynamo still exist in the formstack account they are expected to be.
- If a submission is not found in account 2 we look for it in account 1 before giving up
- every time we fetch a submission from formstack we double check the submission belongs to the email in the original request

### DIFF
--- a/formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/FormstackService.scala
+++ b/formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/FormstackService.scala
@@ -14,16 +14,14 @@ import scalaj.http.{BaseHttp, Http, HttpResponse}
 trait FormstackRequestService {
   def accountFormsForGivenPage(page: Int, accountToken: FormstackAccountToken): Either[Throwable, FormsResponse]
   def formSubmissionsForGivenPage(page: Int, formId: String, minTime: SubmissionTableUpdateDate, encryptionPassword: String, accountToken: FormstackAccountToken): Either[Throwable, FormSubmissions]
-  def submissionData(submissionIdEmails: List[SubmissionIdEmail], config: PerformLambdaConfig): Either[Throwable, List[FormstackSubmissionQuestionAnswer]]
-  def deleteUserData(submissionIdEmails: List[SubmissionIdEmail], config: PerformLambdaConfig): Either[Throwable, List[SubmissionDeletionReponse]]
+  def submissionData(requestEmail:String, submissionIdEmails: List[SubmissionIdEmail], config: PerformLambdaConfig): Either[Throwable, List[FormstackSubmissionQuestionAnswer]]
+  def deleteUserData(requestEmail:String, submissionIdEmails: List[SubmissionIdEmail], config: PerformLambdaConfig): Either[Throwable, List[SubmissionDeletionReponse]]
 }
-
-case class SubmissionsResponse(found: List[Submission], notFound: List[SubmissionIdEmail])
-sealed trait getSubmissionResult
-case class Found(submission:Submission) extends getSubmissionResult
-case class Skipped(submissionIdEmail: SubmissionIdEmail) extends getSubmissionResult
-case class AccountQuestionsAnswersResult(found: List[FormstackSubmissionQuestionAnswer], notFound: List[SubmissionIdEmail] )
-case class ValidatedSubmissions(accountOneResponse:SubmissionsResponse, accountTwoResponse: SubmissionsResponse)
+//this used to group results to multiple formstack calls in found and not found results.
+//for not found results we basically have just the id and for the found results we have whatever the particular api call would return
+case class FormstackResponses[T](found:List[T], notFound: List[SubmissionIdEmail])
+//this just used to group results of fetching submissions by which account they were fetched from. This might not agree with the account number dynamodb might assign to the submission
+case class ValidatedSubmissions(accountOneResponse:FormstackResponses[Submission], accountTwoResponse: FormstackResponses[Submission])
 
 sealed trait FormstackSkippableError extends Throwable
 case class FormstackDecryptionError(message: String) extends FormstackSkippableError
@@ -89,7 +87,7 @@ import FormstackService._
       case _ => decode[FormSubmissions](response.body)
     }
   }
-  
+
   val skippableErrorMessages = List(
     "A valid submission id was not supplied",
     "A valid form could not be found"
@@ -106,14 +104,18 @@ import FormstackService._
   }
 
   def validateEmail(expectedEmail:String, submission: Submission): Boolean = submission.data.exists{
-    subData => extractEmails(subData.value).exists(e => e.equalsIgnoreCase(expectedEmail))
+    subData => {
+      extractEmails(subData.value).exists(e => e.equalsIgnoreCase(expectedEmail))
+    }
   }
 
   private def getSubmissions(
+    requestEmail: String,
     submissionIdEmails: List[SubmissionIdEmail],
     accountToken: FormstackAccountToken,
-    encryptionPassword: String): Either[Throwable, SubmissionsResponse] = {
-    val submissionResults: Either[Throwable, List[getSubmissionResult]] = submissionIdEmails.traverse { submissionIdEmail =>
+    encryptionPassword: String): Either[Throwable, FormstackResponses[Submission]] = {
+
+    val submissionResults: Either[Throwable, List[Either[SubmissionIdEmail, Submission]]] = submissionIdEmails.traverse { submissionIdEmail =>
       val response =
         http(s"https://www.formstack.com/api/v2/submission/${submissionIdEmail.submissionId}.json")
           .header("Authorization", accountToken.secret)
@@ -125,22 +127,22 @@ import FormstackService._
       }
 
       decodeIfNotSkippableError[Submission](response).map{
-        case None => Skipped(submissionIdEmail)
+        case None => Left(submissionIdEmail)
         case Some(submission) =>
           //validate the submission we found in formstack actually contains references to the email we were looking for
-          if (validateEmail(submissionIdEmail.email, submission))
-            Found(submission)
+          if (validateEmail(requestEmail, submission))
+            Right(submission)
           else {
             logger.warn(s"found submission by id = ${submissionIdEmail.submissionId} but the requested email was not found, skipping")
-            Skipped(submissionIdEmail)
+            Left(submissionIdEmail)
           }
       }
     }
 
     submissionResults.map{ subResults =>
-      SubmissionsResponse(
-        found = subResults.collect{ case f:Found => f.submission},
-        notFound = subResults.collect{ case s:Skipped => s.submissionIdEmail}
+      FormstackResponses[Submission](
+        found = subResults.collect{ case Right(submission) => submission},
+        notFound = subResults.collect{ case Left(skippedSubmissionIdEmail) => skippedSubmissionIdEmail}
       )
     }
   }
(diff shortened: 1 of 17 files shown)


formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/FormstackService.scala
```diff
@@ -79,24 +79,33 @@ object FormstackService extends FormstackRequestService with LazyLogging {
...
           .asString
 
-      if(!response.is2xx) {
-        logger.error(response.body)
-      }
+      def isErrorResponse = !response.is2xx
+
+      // Define errors that we can safely skip
+      def isDecryptionError = response.body.contains("An error occurred while decrypting the submissions")
+      def isSubmissionDeletedError = response.body.contains("An error occurred while decrypting the submissions")
```
pvighi: this is probably the wrong error message for this, right ?

pvighi:
[Make baton-formstack handle form deletions](https://trello.com/c/ZQmesRDb/3890-make-baton-formstack-handle-form-deletions)

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/FormstackService.scala
kenoir: Docs seem to [indicate that it'll 404 in this case](https://developers.formstack.com/reference/submission-id-get), and the error will look like:

```json
{
  "status": "error",
  "error": "A valid submission id was not supplied"
}
```

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/FormstackService.scala
```diff
@@ -79,24 +79,33 @@ object FormstackService extends FormstackRequestService with LazyLogging {
...
+      def isFormNotFoundError = response.body.contains("A valid form could not be found")
 
-      /* There are a couple of forms that the Formstack API can't seem to decrypt. There seems to be no way around this
-      *  so we capture this specific error and skip these forms. */
-      if(response.body.contains("An error occurred while decrypting the submission"))
-        Left(FormstackDecryptionError(s"${response.body} | submission id: ${submissionIdEmail.submissionId}"))
-      else
-        decode[Submission](response.body)
+      def isSkippableError = isErrorResponse && isDecryptionError && isSubmissionDeletedError && isFormNotFoundError
```
pvighi: It should be one of the skippable errors, not all at the same time
```suggestion
      def isSkippableError = isErrorResponse && (isDecryptionError || isSubmissionDeletedError || isFormNotFoundError)
```

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/FormstackService.scala
```diff
@@ -137,7 +146,7 @@ object FormstackService extends FormstackRequestService with LazyLogging {
   override def deleteUserData(submissionIdEmails: List[SubmissionIdEmail], config: PerformLambdaConfig): Either[Throwable, List[SubmissionDeletionReponse]] = {
```
pvighi: I can't really think of  a way to test deleting data as it seems we only have a PROD environment. I tried to share as much code with `getSubmissions` to at least have some testing on it

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/FormstackService.scala
```diff
@@ -74,30 +75,38 @@ object FormstackService extends FormstackRequestService with LazyLogging {
     else
       decode[FormSubmissions](response.body)
   }
-
+  
+  val skippableErrorMessages = List(
+    "An error occurred while decrypting the submissions",
```
pvighi: @ob6160  I don't remember why this was added ( I think it was a previous issue) but are we sure we want to skip this error ?

formstack-baton-requests/src/local/com/gu/identity/formstackbatonrequests/services/FormstackRequestServiceLocalRun.scala
```diff
@@ -0,0 +1,36 @@
...
+    submissionId = "1234",
+    receivedByLambdaTimestamp = 0,
+    accountNumber = 1
+  )
+  val submissionIdEmails = List(invalidSubmission)
+  val response = FormstackService.submissionData(submissionIdEmails, config)
+  println("response was:")
+  println(response)
+
+}
```
pvighi: I added this quick test that calls the part of the code that I want to modify. Unfortunately we can only do this for the  function used to process SARs, the RER one has a very similar change but I cannot run this without actually deleting something from PROD.

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/FormstackService.scala
```diff
@@ -74,30 +75,38 @@ object FormstackService extends FormstackRequestService with LazyLogging {
...
+    case Left(e) => Left(e)
+    case Right(submission) => Right(Some(submission))
+  }
+  
   private def getSubmissions(
     submissionIdEmails: List[SubmissionIdEmail],
     accountToken: FormstackAccountToken,
     encryptionPassword: String): Either[Throwable, List[Submission]] = {
-    submissionIdEmails.traverse { submissionIdEmail =>
+    val submissionResults: Either[Throwable, List[Option[Submission]]] = submissionIdEmails.traverse { submissionIdEmail =>
```
kenoir, line 99: Should this be a `Try` if `Left` is always `Throwable`?

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/FormstackService.scala
kenoir: No, we're not. It's not clear in what situations this error would arise.

kenoir: For this PR though we could keep it the same, to avoid changing current behaviour.

pvighi: I think the current behaviour might be wrong..
It catches the "decoding" error but then it still returns a `Left` with a custom error that doesn't get any special treatment later so it still fails the whole thing, just with a different error:
https://github.com/guardian/identity-processes/pull/90/files

pvighi: I think it was modelled after the behaviour when dynamodb is updated with new form data but I think that case it works because:

1. The function that returns the `Left` only makes a single api call so it's fine to return a `Left`. In this one we are in a loop so we want to continue with the next submission and  return a `Right` with a list of all the successful submissions unless there's an unskippable error in which case we would want to return a `Left`
2. The function that calls that has a special case to deal with the custom `FormstackDecryptionError` and not return an error. See [Here](https://github.com/guardian/identity-processes/blob/e92e3ae5becc2419769d35159cb3924c8a23afaa/formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/DynamoUpdateService.scala#L39)

pvighi: In short I think we should probably not skip this error. What do you think @ob6160 ?

pvighi:
```suggestion
```
In short I think we should probably not skip this error. What do you think @ob6160 ?

ob6160: okay, let's not skip it - if it turns out that it breaks things it's not a big deal to add it back in

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/FormstackService.scala
ob6160: I can't either, the best suggestion I've heard so far is to set up a test account in Formstack and run tests against that locally perhaps

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/FormstackService.scala
ob6160: i think we can remove it from the list because on line 73 we have the check from before we came to the codebase: https://github.com/guardian/identity-processes/pull/92/files#diff-2e97f551ddedcd9a50db834124cfb109ad0dc4f748e72de9f9b84623d8625f0bR73

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/FormstackService.scala
```diff
@@ -74,30 +75,41 @@ object FormstackService extends FormstackRequestService with LazyLogging {
...
-
+          
       if(!response.is2xx) {
         logger.error(response.body)
       }
-
-      /* There are a couple of forms that the Formstack API can't seem to decrypt. There seems to be no way around this
-      *  so we capture this specific error and skip these forms. */
-      if(response.body.contains("An error occurred while decrypting the submission"))
-        Left(FormstackDecryptionError(s"${response.body} | submission id: ${submissionIdEmail.submissionId}"))
```
pvighi, line 103: just to sum up our previous conversation for future reference: 

Going from the comment in the code  it seems like the intention was to skip this error and keep going, but the current code wasn't doing doing that, it was just returning a custom error type that nothing picked up later.
We don't think this error should be skipped anyway, but if we later decide it should be we can add it to the list of skippable errors

pvighi:
[formstack migration baton changes](https://trello.com/c/r6zXHbjp/4180-formstack-migration-baton-changes)

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/FormstackService.scala
```diff
@@ -129,25 +170,74 @@ object FormstackService extends FormstackRequestService with LazyLogging {
...
+  override def deleteUserData(requestEmail: String, submissionIdEmails: List[SubmissionIdEmail], config: PerformLambdaConfig): Either[Throwable, List[SubmissionDeletionReponse]] = {
+    for {
+      //We "fix" the submissions that referer to the wrong formstack account in dynamo
+      fixedSubmissionIdEmails <- validateAndFixSubmissionIdEmails(requestEmail, submissionIdEmails, config)
+      deleteResponse <- deleteValidatedSubmissions(fixedSubmissionIdEmails, config)
+    } yield deleteResponse
+
+  }
+
+  def deleteValidatedSubmissions(submissionIdEmails: List[SubmissionIdEmail], config: PerformLambdaConfig): Either[Throwable, List[SubmissionDeletionReponse]] = {
```
pvighi, line 234: the idea of this is that since we cannot test deletions we are just keeping the existing deletion code intact and putting some validation phase before that we can test. Now this should recieve a list of submissions that are supposed to be "fixed": submissions that refer to forms that don't exist are removed,  if dynamo has the wrong account id for a submission it is fixed and if any submission doesn't contain a reference to the original email in the request it is also removed

kelvin-chappell (approved):
It's difficult to follow the whole thing but the latest changes look fine to me 👍 
If we have to maintain this code in the long term we'll need to make it, and the environment, a bit easier to work with.

PR #97, opened 2022-06-07
Add skip for password protected forms

ob6160:
## What does this change?
Some forms return the following 401 response when querying for submissions with the correct credentials.

The query looks like this: `https://www.formstack.com/api/v2/form/$formId/submission.json` and the following is returned from Formstack:

```
{
  "status": "error", 
  "error": "Incorrect password"
}
```

This is undocumented in the Formstack API ( https://developers.formstack.com/reference/form-id-submission-get ) and there is no clear reason as to why this is happening; so we skip this response in a similar way to how we were already skipping responses with a decryption error.

### DIFF
--- a/formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/FormstackService.scala
+++ b/formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/FormstackService.scala
@@ -69,7 +69,8 @@ object FormstackService extends FormstackRequestService with LazyLogging {
 
     /* There are a couple of forms that the Formstack API can't seem to decrypt. There seems to be no way around this
      *  so we capture this specific error and skip these forms. */
-    if(response.body.contains("An error occurred while decrypting the submissions"))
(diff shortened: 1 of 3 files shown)


formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/FormstackService.scala
```diff
@@ -69,7 +69,8 @@ object FormstackService extends FormstackRequestService with LazyLogging {
 
     /* There are a couple of forms that the Formstack API can't seem to decrypt. There seems to be no way around this
      *  so we capture this specific error and skip these forms. */
-    if(response.body.contains("An error occurred while decrypting the submissions"))
+    // We also skip responses from Formstack that contain "Incorrect password". These are undocumented, and we do not know why they happen.
```
kenoir, line 74: Can we identify which forms these are and `WARN` log the ID & reason here (same for the decryption errors)?

I think this is an ok fix, so long as these auth errors are happening extremely infrequently.

shtukas (approved):
LGTM

PR #101, opened 2022-09-28
Change holidays to guardian traveller

dblatcher:
## What does this change?

 - Changes the hook for formstack form #1945214 to trigger a subscription for "Guardian Traveller" (editorial newsletter with identityName "guardian-traveller") instead of the consent-based email "Guardian Holidays", which traveller is replacing.
 - Updates the test for 'set-consent' request to use a different Newsletter (was using `Holidays`).
 - Updates the readme about the usage of this feature.

After this PR is merged, the form on https://holidays.theguardian.com/newsletter/ should trigger subscriptions to Guardian Traveller - I think this is the only place the formstack-consents feature is used for this form id.

## Have we considered potential risks?
As with  [this pr in identity](https://github.com/guardian/identity/pull/2148), the timeing of the merge needs to be after "holidays" is replaced by "guardian traveller"

### DIFF
--- a/formstack-consents/README.md
+++ b/formstack-consents/README.md
@@ -4,9 +4,9 @@ Upon a submission to one of the [Formstack](https://guardiannewsandmedia.formsta
 
 When the user clicks the link in the email, an identity account will be created for them and they will be signed up to the newsletter.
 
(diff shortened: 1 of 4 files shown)


kelvin-chappell (approved):
👍

PR #102, opened 2022-10-17
Pv/formstack baton error handling

pvighi:
This PR was based on https://github.com/guardian/identity-processes/pull/92 which has not been merged yet. That pr changed the process so that it doesn't fail if a submission is not found in formstack.

## What does this change?
Forms are going to be migrated from account 2 to account 1, we need a way to find submissions that originally were in account 2 but have been moved to account 1.
main changes

1. made formstackService testable avoiding big refactors as we don't have a code instance to test this
2. added unit test for formstackservice
3. added validation so that every time we retrieve a submission from Formstack we double check the email to make sure it mentions the original email in the request
4. changed the SAR process so that any form 2 submissions that are not found we look for in account 1 in case they have been migrated there.
5.  Added a preparation step to the RER code so that we validate all the submission ids in the request to make sure all submissions exist and they are in the expected account.  If we find any submissions have been migrated  from account 2 to 1 we try to delete them from there instead

### DIFF
--- a/formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/rer/FormstackPerformRerHandler.scala
+++ b/formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/rer/FormstackPerformRerHandler.scala
@@ -19,7 +19,7 @@ case class FormstackPerformRerHandler(
   def initiateRer(request: RerPerformRequest): Either[Throwable, S3WriteSuccess] =
     for {
       submissionIds <- dynamoClient.userSubmissions(request.subjectEmail.toLowerCase, config.bcryptSalt, config.submissionTableName)
-      _ <- formstackClient.deleteUserData(submissionIds, config)
+      _ <- formstackClient.deleteUserData(request.subjectEmail.toLowerCase, submissionIds, config)
       _ <- dynamoClient.deleteUserSubmissions(submissionIds, config.bcryptSalt, config.submissionTableName)
       writeToS3Response <- s3Client.writeSuccessResult(request.initiationReference, List.empty, RER, config)
     } yield writeToS3Response
--- a/formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/FormstackService.scala
+++ b/formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/FormstackService.scala
@@ -14,16 +14,14 @@ import scalaj.http.{BaseHttp, Http, HttpResponse}
 trait FormstackRequestService {
   def accountFormsForGivenPage(page: Int, accountToken: FormstackAccountToken): Either[Throwable, FormsResponse]
   def formSubmissionsForGivenPage(page: Int, formId: String, minTime: SubmissionTableUpdateDate, encryptionPassword: String, accountToken: FormstackAccountToken): Either[Throwable, FormSubmissions]
-  def submissionData(submissionIdEmails: List[SubmissionIdEmail], config: PerformLambdaConfig): Either[Throwable, List[FormstackSubmissionQuestionAnswer]]
-  def deleteUserData(submissionIdEmails: List[SubmissionIdEmail], config: PerformLambdaConfig): Either[Throwable, List[SubmissionDeletionReponse]]
+  def submissionData(requestEmail:String, submissionIdEmails: List[SubmissionIdEmail], config: PerformLambdaConfig): Either[Throwable, List[FormstackSubmissionQuestionAnswer]]
+  def deleteUserData(requestEmail:String, submissionIdEmails: List[SubmissionIdEmail], config: PerformLambdaConfig): Either[Throwable, List[SubmissionDeletionReponse]]
 }
-
-case class SubmissionsResponse(found: List[Submission], notFound: List[SubmissionIdEmail])
-sealed trait getSubmissionResult
-case class Found(submission:Submission) extends getSubmissionResult
-case class Skipped(submissionIdEmail: SubmissionIdEmail) extends getSubmissionResult
-case class AccountQuestionsAnswersResult(found: List[FormstackSubmissionQuestionAnswer], notFound: List[SubmissionIdEmail] )
-case class ValidatedSubmissions(accountOneResponse:SubmissionsResponse, accountTwoResponse: SubmissionsResponse)
+//this used to group results to multiple formstack calls in found and not found results.
+//for not found results we basically have just the id and for the found results we have whatever the particular api call would return
+case class FormstackResponses[T](found:List[T], notFound: List[SubmissionIdEmail])
+//this just used to group results of fetching submissions by which account they were fetched from. This might not agree with the account number dynamodb might assign to the submission
+case class ValidatedSubmissions(accountOneResponse:FormstackResponses[Submission], accountTwoResponse: FormstackResponses[Submission])
 
 sealed trait FormstackSkippableError extends Throwable
 case class FormstackDecryptionError(message: String) extends FormstackSkippableError
@@ -89,7 +87,7 @@ import FormstackService._
       case _ => decode[FormSubmissions](response.body)
     }
   }
-  
+
   val skippableErrorMessages = List(
     "A valid submission id was not supplied",
     "A valid form could not be found"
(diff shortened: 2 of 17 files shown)


formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/FormstackService.scala
```diff
@@ -97,26 +105,48 @@ object FormstackService extends FormstackRequestService with LazyLogging {
...
       if(!response.is2xx) {
         logger.error(response.body)
       }
-      
-          decodeIfNotSkippableError[Submission](response)
+
+      decodeIfNotSkippableError[Submission](response).map{
+        case None => Skipped(submissionIdEmail)
+        case Some(submission) =>
+          //validate the submission we found in formstack actually contains references to the email we were looking for
```
pvighi, line 132: this is always useful but in particular now that we are going to be looking at submissions on different formstack accounts than the one they were originally discovered in we really want to make extra sure it's not a completely unrelated submission that happens to have the same id. This is probably unlikely to happen but it's good to not have to worry about it

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/FormstackService.scala
```diff
@@ -141,25 +171,73 @@ object FormstackService extends FormstackRequestService with LazyLogging {
...
+    }
+    for {
+      validatedSubmissions <- getValidatedSubmissionData(submissionIdEmails, config)
+    } yield fixedSubmissionsFor(validatedSubmissions.accountOneResponse, 1) ++ fixedSubmissionsFor(validatedSubmissions.accountTwoResponse, 2)
+  }
   override def deleteUserData(submissionIdEmails: List[SubmissionIdEmail], config: PerformLambdaConfig): Either[Throwable, List[SubmissionDeletionReponse]] = {
+    for {
+      //we regenerate the submissionIdEmails to make sure they all refer to the formstack account where the submission was verified to exist
+      fixedSubmissionIdEmails <- validateAndFixSubmissionIdEmails(submissionIdEmails, config)
+      deleteResponse <- deleteValidatedSubmissions(fixedSubmissionIdEmails, config)
```
pvighi, line 229: `deleteValidatedSubmissions` is just a rename of the current deleteUserData function.
The idea is to leave the deletion code as unchanged as possible since we cannot test it. This just adds a step at the start to "fix" the list of submissions that we are going to delete. We remove the submissions that are not found anywhere and we change the account id of the submissions that are now in account 1 even though dynamodb says they should be in account 2

formstack-baton-requests/src/test/scala/com/gu/identity/formstackbatonrequests/services/FormstackServiceSpec.scala
```diff
@@ -0,0 +1,398 @@
...
+
+  /**
+   * The library this project uses has a builder pattern where you first get a request object, then you set everything on it and finally call a method like toString to execute it.
+   * Ideally we would have just refactored it so that the actual lower level http stuff is in a client class separate from the service but for now we are mocking this the best we can.
+   * this function returns a fake request that expects the correct things to be set on it and returns a fake response. It doesn't go as far as to check that everything is called in the
+   * right sequence though
+   * @param responseBody
+   * @param status
+   * @param token
+   * @return
```
kelvin-chappell, line 137: Could miss these out if not defined

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/FormstackService.scala
```diff
@@ -17,17 +19,23 @@ trait FormstackRequestService {
   def deleteUserData(submissionIdEmails: List[SubmissionIdEmail], config: PerformLambdaConfig): Either[Throwable, List[SubmissionDeletionReponse]]
 }
 
+case class SubmissionsResponse(found: List[Submission], notFound: List[SubmissionIdEmail])
+sealed trait getSubmissionResult
+case class Found(submission:Submission) extends getSubmissionResult
+case class Skipped(submissionIdEmail: SubmissionIdEmail) extends getSubmissionResult
+case class AccountQuestionsAnswersResult(found: List[FormstackSubmissionQuestionAnswer], notFound: List[SubmissionIdEmail] )
+case class ValidatedSubmissions(accountOneResponse:SubmissionsResponse, accountTwoResponse: SubmissionsResponse)
```
kelvin-chappell: Perhaps some documentation around these types would make it easier to get into the code.

kelvin-chappell (approved):
The logic of it looks fine.  It's inevitably complicated given the nature of the transition.  Perhaps transition-specific code could be marked in some way unless that makes it more complicated. 👍

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/FormstackService.scala
kenoir: Or perhaps also moving them to another file?

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/FormstackService.scala
```diff
@@ -17,17 +18,23 @@ trait FormstackRequestService {
...
+
 sealed trait FormstackSkippableError extends Throwable
 case class FormstackDecryptionError(message: String) extends FormstackSkippableError
 case class FormstackAuthError(message: String) extends FormstackSkippableError
 
-object FormstackService extends FormstackRequestService with LazyLogging {
+case class FormstackService(http:BaseHttp = Http) extends FormstackRequestService with LazyLogging {
```
kenoir: Would this make more sense as a plain `class`?

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/FormstackService.scala
pvighi: yeah sorry I'll try to clean up. I forgot to go over the code before opening the PR to make it a bit more readable.  I got distracted with some unrelated issues and then I forgot to do it

formstack-baton-requests/src/test/scala/com/gu/identity/formstackbatonrequests/services/FormstackServiceSpec.scala
```diff
@@ -0,0 +1,503 @@
+package com.gu.identity.formstackbatonrequests.services
```
kenoir, line 1: The addition of these tests is great.

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/FormstackService.scala
```diff
@@ -125,22 +127,22 @@ import FormstackService._
...
 
       decodeIfNotSkippableError[Submission](response).map{
-        case None => Skipped(submissionIdEmail)
+        case None => Left(submissionIdEmail)
         case Some(submission) =>
           //validate the submission we found in formstack actually contains references to the email we were looking for
-          if (validateEmail(submissionIdEmail.email, submission))
-            Found(submission)
+          if (validateEmail(requestEmail, submission))
+            Right(submission)
```
pvighi, line 134: the bug I found was that this `submissionIdEmail.email` is actually a one way hash of the email as stored in the db, not the actual email.
The solution I came up with when I found this bug late yesterday was to pass around the original request email along with all the submission ids we got from the db and use that to validate that everything we get from formstack contains that email. Another option would have been to pass around the dependencies we need to do the hash and hash what we get from formstack and compare the two hashes instead

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/rer/FormstackPerformRerHandler.scala
```diff
@@ -19,7 +19,7 @@ case class FormstackPerformRerHandler(
   def initiateRer(request: RerPerformRequest): Either[Throwable, S3WriteSuccess] =
     for {
       submissionIds <- dynamoClient.userSubmissions(request.subjectEmail.toLowerCase, config.bcryptSalt, config.submissionTableName)
-      _ <- formstackClient.deleteUserData(submissionIds, config)
+      _ <- formstackClient.deleteUserData(request.subjectEmail.toLowerCase, submissionIds, config)
```
pvighi, line 22: see comment below, but basically the the `email` field of the submissionsIds doesn't contain the email as I was assuming, it actually has a hash of the email, so I pass the actual email in the request to use that instead

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/FormstackService.scala
```diff
@@ -14,16 +14,14 @@ import scalaj.http.{BaseHttp, Http, HttpResponse}
...
-sealed trait getSubmissionResult
-case class Found(submission:Submission) extends getSubmissionResult
-case class Skipped(submissionIdEmail: SubmissionIdEmail) extends getSubmissionResult
-case class AccountQuestionsAnswersResult(found: List[FormstackSubmissionQuestionAnswer], notFound: List[SubmissionIdEmail] )
-case class ValidatedSubmissions(accountOneResponse:SubmissionsResponse, accountTwoResponse: SubmissionsResponse)
+//this used to group results to multiple formstack calls in found and not found results.
+//for not found results we basically have just the id and for the found results we have whatever the particular api call would return
+case class FormstackResponses[T](found:List[T], notFound: List[SubmissionIdEmail])
+//this just used to group results of fetching submissions by which account they were fetched from. This might not agree with the account number dynamodb might assign to the submission
+case class ValidatedSubmissions(accountOneResponse:FormstackResponses[Submission], accountTwoResponse: FormstackResponses[Submission])
```
pvighi, line 24: I cleared out some of these classes. The `getSubmissionResult` I ended up removing in favour of just using either as it was only used inside a single function one time.
The two case classes that basically separate results in `found` and `not found` I just refactored them into a single generic class

pvighi:
[formstack migration baton changes](https://trello.com/c/r6zXHbjp/4180-formstack-migration-baton-changes)

PR #103, opened 2022-10-26
Remove all PROD references from formstack-baton-requests CODE 

pvighi:
## What does this change?
The `CODE` instance of formstack-baton-requests was configured with all `PROD` credentials with only the last step of the SAR or RER being replaced with a stub that returned a fake result to Baton. 

This meant that every time Baton was executed in `CODE` formstack-baton-requests would fetch the latests submissions from Formstack and update the production dynamodb table. I don't think this broke anything because it was the exact same code as `PROD` but it is unsafe.

**Change made outside this PR:**
_ Edited the cloudformation parameters in `CODE`  to not refer to any production resources:
 - new dynamodb table names and created a different bcrypt salt
 -  created a `no data` parameter to pass in place of the formstack credentials. This is just to keep things from failing in CODE, but in this PR we are replacing the formstack service with a stub as we still don't have a safe formstack account to use for testing

These changes in the cfn parameters mean that not only will the step function not attempt to access the production resources, but it will not even have the permissions to do so

### DIFF
--- a/formstack-baton-requests/cloud-formation.yaml
+++ b/formstack-baton-requests/cloud-formation.yaml
@@ -154,10 +154,7 @@ Resources:
       Role:
         !GetAtt FormstackBatonSarLambdaRole.Arn
 
-#  There tables are only created in PROD since Formstack doesn't have a CODE environment and we don't want to test in
-#  CODE with real data
   FormstackSubmissionIds:
-    Condition: IsProd
     Type: AWS::DynamoDB::Table
     Properties:
       TableName: !Ref SubmissionsTableName
@@ -176,7 +173,6 @@ Resources:
         WriteCapacityUnits: 20
 
   FormstackSubmissionsLastUpdated:
-    Condition: IsProd
     Type: AWS::DynamoDB::Table
     Properties:
       TableName: !Ref LastUpdatedTableName
--- a/formstack-baton-requests/cloud-formation.yaml
+++ b/formstack-baton-requests/cloud-formation.yaml
@@ -20,22 +20,17 @@ Parameters:
     Type: CommaDelimitedList
   EncryptionPasswordPath:
     Type: String
-    Default: /identity/formstack-baton-requests/encryption-password
   FormstackAccountOneTokenPath:
     Type: String
-    Default: /identity/formstack-baton-requests/formstack-account-one-token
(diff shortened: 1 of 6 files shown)


formstack-baton-requests/cloud-formation.yaml
```diff
@@ -20,22 +20,17 @@ Parameters:
     Type: CommaDelimitedList
   EncryptionPasswordPath:
     Type: String
-    Default: /identity/formstack-baton-requests/encryption-password
```
pvighi, line 23: removed the defaults to ensure we are intentionally loading the right credentials for the stage

formstack-baton-requests/cloud-formation.yaml
```diff
@@ -154,10 +149,7 @@ Resources:
       Role:
         !GetAtt FormstackBatonSarLambdaRole.Arn
 
-#  There tables are only created in PROD since Formstack doesn't have a CODE environment and we don't want to test in
-#  CODE with real data
```
pvighi, line 158: I changed  the table names in the `CODE` stack params so we can let cfn create them

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/Handler.scala
```diff
@@ -73,13 +74,16 @@ trait FormstackHandler[Req, Res] extends LazyLogging {
...
 
   val stage: String = sys.env.getOrElse("STAGE", "CODE") match {
     case "PROD" => "PROD"
     case _      => "CODE"
   }
-  val formstackService = new FormstackService()
+  def formstackService = if (stage == "PROD") new FormstackService() else {
+    logger.info(s"using mocked formstack service as stage is $stage")
+    FormstackServiceStub.withSuccessResponse
+  }
```
pvighi, line 86: this is not the best, but until we get some sort of Formstack sandbox environment we are just using the stub we had for tests in non `PROD` environments

Mark-McCracken (approved):
That looks all good to me, very nice

pvighi:
[formstack-baton-requests - remove PROD references from CODE instance](https://trello.com/c/NhuvdkH6/4203-formstack-baton-requests-remove-prod-references-from-code-instance)

PR #104, opened 2022-10-27
Pv/remove formstack baton remove account two

pvighi:
## What does this change?
Formstack forms are being migrated away from account 2.  The current code assumes we have 2 formstack accounts and will cause failures is the credentials for account 2 stop working. 
This PR removes account 2 references and makes the project work with a single account.
There's still more refactoring that could be done to simplify the code now that there's a single account but we will leave that for later, maybe once we get a fully functional `CODE`  environment.

### before
image

### after
image

### before
image

### after
image

### DIFF
--- a/formstack-baton-requests/cloud-formation.yaml
+++ b/formstack-baton-requests/cloud-formation.yaml
@@ -646,79 +646,32 @@ Resources:
         - |-
           {
             "Comment": "A step function for performing Formstack RERs",
-            "StartAt": "UpdateDynamoInParallel",
+            "StartAt": "AddAccountOneParam",
             "States": {
-              "UpdateDynamoInParallel": {
-                "Type": "Parallel",
-                "Next": "PerformRer",
-                "Branches": [
+              "AddAccountOneParam": {
+                "Type": "Pass",
+                "Next": "UpdateDynamoWithAccountOne",
+                "Result": "1",
+                "ResultPath": "$.accountNumber"
+              },
+              "UpdateDynamoWithAccountOne": {
+                "Type": "Task",
+                "Resource": "${updateLambdaArn}",
+                "Next": "CheckAccountOneUpdateCompletion"
+              },
+              "CheckAccountOneUpdateCompletion": {
+                "Type": "Choice",
+                "Choices": [
                   {
-                    "StartAt": "AddAccountOneParam",
-                    "States": {
-                      "AddAccountOneParam": {
-                        "Type": "Pass",
-                        "Next": "UpdateDynamoWithAccountOne",
-                        "Result": "1",
-                        "ResultPath": "$.accountNumber"
-                      },
-                      "UpdateDynamoWithAccountOne": {
-                        "Type": "Task",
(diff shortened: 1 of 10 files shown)


pvighi:
[Formstack-baton-requests - make changes to use only 1 account](https://trello.com/c/IsWdIjIW/4201-formstack-baton-requests-make-changes-to-use-only-1-account)

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/updatedynamo/UpdateDynamoHandler.scala
```diff
@@ -19,33 +18,30 @@ case class UpdateDynamoHandler(
...
       } else Right(UpdateStatus(completed = true, None, None, token))
     }
   }
 
   override def handle(request: UpdateDynamoRequest, context: Context): Either[Throwable, UpdateDynamoResponse] = {
-    val accountNumber = request.accountNumber match {
-      case Some(account) => account
+    val token = request.accountNumber match {
+      case Some(account) if account == 1 => config.accountOneToken
+      case Some(account) => throw new RuntimeException(s"Unexpected account number: $account")
```
pvighi, line 40: we are leaving this for a future change, maybe after improving the `CODE` env but we could go even further and remove the account parameter from everywhere. 
Currently the requests for the lambdas and the dynamo table have an account parameter field,  the dynamodb table still has the old submissions with their original account number but the code should change it to 1 as it's the only Formstack account to exist once the migration is over.

formstack-baton-requests/src/test/scala/com/gu/identity/formstackbatonrequests/services/FormstackServiceSpec.scala
```diff
@@ -304,106 +276,25 @@ class FormstackServiceSpec extends FreeSpec with Matchers with MockFactory {
       formstackService.submissionData("aDifferentEmailNotInFormastack@email.com", List(submissionIdEmail222InAccountTwo), config) shouldBe Right(List.empty)
          //just to verify the two calls that are made to the same endpoint are done in the correct sequence of first attempting account two and then account one
       inSequence {
-        (submissionNotFoundInAccountTwo.header _).verify("Authorization", config.accountTwoToken.secret)
-        (submissionNotFoundInAccountTwo.asString _).verify()
+  //      (submissionNotFoundInAccountTwo.header _).verify("Authorization", config.accountTwoToken.secret)
+   //     (submissionNotFoundInAccountTwo.asString _).verify()
```
kelvin-chappell, line 280: Remove

kelvin-chappell (approved):
Looks fine 👍

formstack-baton-requests/cloud-formation.yaml
```diff
@@ -600,79 +600,33 @@ Resources:
         - |-
           {
             "Comment": "A step function for performing Formstack SARs",
```
pvighi, line 602: see screenshots in PR description to see the graphical representation of the changes to the step functions

PR #243, opened 2023-07-12
Migrate from TeamCity to GitHub Actions

raphaelkabo:
- Replace existing TeamCity CI behaviour with a GitHub Actions workflow
- Remove the RiffRaffArtifact plugin
- Limit the existing workflow for the formstack consents lambda to only run when its files are updated

## Testing

- [x] Test lambda in DEV

### DIFF
--- a/.github/workflows/ci-eventbrite-consents.yml
+++ b/.github/workflows/ci-eventbrite-consents.yml
@@ -1,0 +1,54 @@
+name: CI - Eventbrite consents sync lambda
+
+on:
+  pull_request:
+    paths:
+      - 'eventbrite-consents/**'
+      - '.github/workflows/ci-eventbrite-consents.yml'
+  push:
+    branches:
+      - main
+    paths:
+      - 'eventbrite-consents/**'
+      - '.github/workflows/ci-eventbrite-consents.yml'
+  workflow_dispatch:
+
+jobs:
+  CI:
+    runs-on: ubuntu-latest
+
+    permissions:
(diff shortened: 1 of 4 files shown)


raphaelkabo:
@akinsola-guardian, I've made a small change to your workflow file so that it only runs when the files in that directory are modified (otherwise it's currently running whenever anything happens to one of the other projects). Could you let me know if there's any issues with that?

.github/workflows/ci-eventbrite-consents.yml
```diff
@@ -0,0 +1,54 @@
...
+
+      - name: Upload to RiffRaff
+        uses: guardian/actions-riff-raff@v2
+        with:
+          projectName: eventbrite-consents-lambda
+          buildNumberOffset: 82
+          configPath: eventbrite-consents/riff-raff.yaml
+          contentDirectories: |
+            eventbrite-consents-lambda:
+              - eventbrite-consents/target/scala-2.12/main.jar
```
Mark-McCracken, line 52: I struggled in getting this to work sometimes, because the scala version was different from what I got locally, so I moved it to the root dir and renamed, if you see anything like this:
```yaml
- run: |
    sbt compile test assembly
    mv target/scala*/stupidly-named.jar ./good-name.jar
```
you'll know why

Mark-McCracken (approved):
Looks great.
I checked the output in the riffraff bucket, looks consistent with older teamcity builds, should work fine

PR #246, opened 2023-08-10
use eastern time on formstack call

pvighi:
## What does this change?

We are currently using the GET /form/:id/submission endpoint in the Formstack APi and providing the `min_time`  parameter to fetch the submissions that happened since the last time this process runs.
The problem is that this api call expects this parameter to be in USA eastern time (apparently not Eastern Standard time, which ignores daylight savings).
Even though the dates are returned in uk time (or maybe in the time zone the form is  configured to use), the filtering itself must be provided in usa eastern time.

See documentation here: https://developers.formstack.com/reference/form-id-submission-get.

### DIFF
--- a/formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/aws/DynamoClient.scala
+++ b/formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/aws/DynamoClient.scala
@@ -1,8 +1,7 @@
 package com.gu.identity.formstackbatonrequests.aws
 
 import java.time.format.DateTimeFormatter
-import java.time.LocalDateTime
-
+import java.time.{LocalDateTime, ZoneId, ZoneOffset}
 import com.typesafe.scalalogging.LazyLogging
 import com.amazonaws.services.dynamodbv2.{AmazonDynamoDB, AmazonDynamoDBClient}
 import com.gu.scanamo.Scanamo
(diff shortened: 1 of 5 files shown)


formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/FormstackService.scala
```diff
@@ -72,7 +72,8 @@ import FormstackService._
           ("data", "true"),
           ("expand_data", "true"),
           ("sort", "DESC"),
-          ("min_time", minTime.date)
+          // this api call expects eastern time zone, see https://developers.formstack.com/reference/form-id-submission-get
+          ("min_time", minTimeUTC.toEasternTime)
```
pvighi, line 76: See documentation here: https://developers.formstack.com/reference/form-id-submission-get.
image

formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/aws/DynamoClient.scala
```diff
@@ -22,7 +22,12 @@ trait DynamoClient {
   def deleteUserSubmissions(submissionIdsAndEmails: List[SubmissionIdEmail], salt: String, submissionsTableName: String): Either[Throwable, List[DeleteItemResult]]
 }
 
-case class SubmissionTableUpdateDate(formstackSubmissionTableMetadata: String, date: String)
+case class SubmissionTableUpdateDate(formstackSubmissionTableMetadata: String, date: String){
+  def toEasternTime:String = {
+    val utcDate = LocalDateTime.parse(date, formatter).atZone(ZoneOffset.UTC)
+    utcDate.withZoneSameInstant(ZoneId.of("America/New_York")).format(formatter)
+  }
```
pvighi, line 29: I think this date is just a string because it's read from dynamodb and then directly used to pass in to formstack but now we have to parse it into a date to change the timezone and then parse change it back into a string.

Another alternative would be to always use eastern time everywhere but I thought it is expected that dates are in UTC so maybe it's less confusing to translate it into eastern just when we pass it to formstack but leave it as UTC everywhere else

coldlink (dismissed):
timezones should be abolished

kelvin-chappell (approved):
👍

PR #247, opened 2023-08-14
formstack baton request gha fixes

pvighi:
The main reason for this pr is that the continuous deployment is not triggering because the project name has changed.
Instead of changing the expected project name on riff-raff we will fix it here to match the rest of the projects in the repo.
before :
image

after:
image

I also made a couple other changes to match the other projects in the repo:
- add a manual trigger for the build 
- give the project a name

It looks like the manual trigger might not show up until you merge the pr?

### DIFF
--- a/.github/workflows/deploy-formstack.yml
+++ b/.github/workflows/deploy-formstack.yml
@@ -1,3 +1,4 @@
+name: CI - Formstack Baton Requests
 on:
   pull_request:
     paths:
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-formstack.yml
```diff
@@ -9,6 +10,7 @@ on:
     paths:
       - 'formstack-baton-requests/**'
       - '.github/workflows/deploy-formstack.yml'
+  workflow_dispatch:
```
pvighi, line 13: it looks like this manual trigger button doesn't show up until you merge the PR?

akinsola-guardian (approved):
LGTM. Thanks for fixing this

raphaelkabo (approved):
Looks great.

PR #248, opened 2023-08-14
update error message on skippable error

pvighi:
## What does this change?
it looks like the error message we get when we provide the wrong encryption password changed. The code looks at the error message to decide whether it should skip the error or not

### DIFF
--- a/formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/FormstackService.scala
+++ b/formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/FormstackService.scala
@@ -83,12 +83,12 @@ import FormstackService._
 
     /* There are a couple of forms that the Formstack API can't seem to decrypt. There seems to be no way around this
      *  so we capture this specific error and skip these forms. */
-    // We also skip responses from Formstack that contain "Incorrect password". These are undocumented, and we do not know why they happen.
+    // We also skip responses from Formstack that contain "Form encryption password invalid". These are undocumented, and we do not know why they happen.
 
     response.body match {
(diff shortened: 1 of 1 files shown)


formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/FormstackService.scala
```diff
@@ -83,12 +83,12 @@ import FormstackService._
...
     /* There are a couple of forms that the Formstack API can't seem to decrypt. There seems to be no way around this
      *  so we capture this specific error and skip these forms. */
-    // We also skip responses from Formstack that contain "Incorrect password". These are undocumented, and we do not know why they happen.
+    // We also skip responses from Formstack that contain "Form encryption password invalid". These are undocumented, and we do not know why they happen.
 
     response.body match {
       case message if message.contains("An error occurred while decrypting the submissions") =>
         Left(FormstackDecryptionError(s"${response.body} | form id: ${formId}"))
-      case message if message.contains("Incorrect password") =>
+      case message if message.contains("Form encryption password invalid") =>
```
pvighi, line 91: looking at the logs we are getting this new error message for the same form id that was returning the old error message (and getting skipped) last week

pvighi: This is just a change to try to get everything working the way it was, but I don't really like this solution because if we ever get the wrong encryption password in our configuration we would silently skip forms that should have been processed.
I'll try to find out why these forms have custom encryption settings but I'm not sure it will help mitigate this risk

coldlink (approved):

PR #249, opened 2023-08-17
update dynamodb settings

pvighi:
## What does this change?
I noticed the code dynamodb table has the same capacity as the prod but it  is barely used. There's no need to pay $11 a month when we can pay $0.60 
Also I enabled deletion protection on the tables just because it's good practice

## Test

I deployed these changes to code and I can see the settings being applied correctly.


kelvin-chappell (dismissed):
Well spotted! 👍

raphaelkabo (approved):

PR #251, opened 2023-09-05
skip ybtj forms

pvighi:
skip forms that often cause api errors but have no relevant data

### DIFF
--- a/formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/DynamoUpdateService.scala
+++ b/formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/DynamoUpdateService.scala
@@ -138,9 +138,16 @@ case class DynamoUpdateService(
       case Left(err) => Left(err)
       case Right(response) =>
         val forms = response.forms
-        val formResults = forms.map { form =>
-          logger.info(s"Processing results for form ${form.id}")
-          writeSubmissions(form, lastUpdate, token = token)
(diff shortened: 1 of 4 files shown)


formstack-baton-requests/src/main/scala/com/gu/identity/formstackbatonrequests/services/DynamoUpdateService.scala
```diff
@@ -138,9 +138,16 @@ case class DynamoUpdateService(
       case Left(err) => Left(err)
       case Right(response) =>
         val forms = response.forms
-        val formResults = forms.map { form =>
-          logger.info(s"Processing results for form ${form.id}")
-          writeSubmissions(form, lastUpdate, token = token)
+        val formResults: Seq[Either[Throwable, Unit]] = forms.map { form =>
+          // The amount of new submissions to these forms often exceeds an apparent formstack limit on how many can be fetched in a single query (even if the api call is paginated)
+          // The submissions would be ignored anyway since the form doesn't have any relevant data, so we just skip them at the form level to avoid the problem for now.
+           if (form.name.toLowerCase.startsWith("ybtj")) {
```
kelvin-chappell, line 144: Perhaps extract this condition into its own method called something like `mightHaveEmailField(form)` and swap the outcomes.

kelvin-chappell (approved):
Seems quite bizarre but looks like it's the only way at the moment! 👍 

Perhaps link to the Trello card on the PR description to give context.

PR #250, opened 2023-08-22
formstack-baton-requests: allow partial dynamo db updates & and add step functions to do just the dynamo update

pvighi:
- add a max_time parameter so that we can update submissions that happened between 2 timestamps.
- add a dedicated state machine just to update dynamodb.

PR #252, opened 2023-09-14
added some documentation for recent formstack-baton-requests changes

pvighi:
## What does this change?
Add some comments and documentation left out of https://github.com/guardian/identity-processes/pull/250

### DIFF
--- a/formstack-baton-requests/README.md
+++ b/formstack-baton-requests/README.md
@@ -19,6 +19,19 @@ The `FormstackRer` step function is triggered by the `FormstackRerHandler` with 
 3. Performs a check to see if the status in the `UpdateDynamoResponse` is `Completed`, if it's not, the `UpdateDynamoHandler` is triggered again to continue the update. If a `Completed` status is found, the update branch is completed for that account's token.
 4. Once updates are completed for both tokens, a the `FormstackPerformRerHandler` is triggered (see more information on this lambda below).
 
+### FormstackUpdateDynamo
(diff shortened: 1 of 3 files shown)


formstack-baton-requests/README.md
```diff
@@ -29,6 +42,8 @@ The UpdateDynamoHandler updates DynamoDb with submissions received by Formstack
 4. After each page of forms returned by the Formstack API have been processed, the UpdateDynamoHandler checks the remaining time the lambda has to run. If it is less than 5 minutes but there are still more forms to be processed, an UpdateDynamoResponse is returned with a `Pending` status. Otherwise, it has a `Completed` status.
 5. Once a `Completed` status is returned, the DynamoDb table storing the last run date is updated for that account.
 
+**Note:** The optional "maxUpdateSeconds" parameter should be left empty for regular SAR and RER operation to ensure the lambda does a full update. This parameter is only meant for use by devs to do partial updates during backfills or troubleshooting issues. For more details see the [FormstackUpdate Dynamo step functions](#formstackupdatedynamo) section of this document.
+
```
pvighi, line 46: I'm mostly worried by the "count" parameter that for some reason has to be initialised to the amount of forms per page in the formstack response. The problem is that this is a constant in Scala code that is not available in from the cloudformation template.
The ideal solution to this would be to remove the need to externally set this initial value or at least make it something that is more intuitive but I'm avoiding refactoring this for now

kelvin-chappell (approved):
Good to know.  Thanks for doing this 👍

PR #254, opened 2023-11-13
increase timeout for formstack calls

pvighi:
## What does this change?
Increase timeouts for formstack api calls. We observed successful api calls taking 20 seconds to return, this PR increases the timeout from 10s to 30s.
For more information see https://github.com/guardian/identity-processes/pull/241 where we increased the same limits in the past.
## How to test
see note about testing in linked PR, we cannot run the whole thing but we can test the code that runs the api call in isolation

the formstack process finishes execution instead of timing out on a formstack call


coldlink (approved):

pvighi:
timeouts are not an issue anymore

PR #265, opened 2024-04-26
put back previous timeout for lambda execution

pvighi:
puts back the lambda remaining execution time limit to what it was before https://github.com/guardian/identity-processes/pull/259.

it helps fewer calls on manual updates and combined with an existing bug that we should address later it makes it harder to catch up the dynamo table to fix baton timeout errors.


kelvin-chappell (approved):
😱

